PR #15, opened 2015-06-05
Docs / job_runner.sh reference previous 0.1.3 jar

benlee:
- I'm not very familiar with gradle, so not sure
  if there's a programmatic way to get this to line up


hectorgon:
Thank you this has been fixed by another pull request, will try to figure out how to do it automatically

Issue #18, opened 2015-06-06
Income prediction demo job_runner.sh fails bc it refers to old library file.


byuksel:
"/aerosolve/demo/income_prediction/sh job_runner.sh MakeTraining" returns:

/aerosolve/demo/income_prediction/build/libs/income_prediction-0.1.3-all.jar does not exist, skipping.
java.lang.ClassNotFoundException: com.airbnb.aerosolve.demo.IncomePrediction.JobRunner

income_prediction-0.1.3 is now income_prediction-0.1.6

PR #19, opened 2015-06-06
Fix issue #18 and #20, update the library version

byuksel:
Hi, 

While trying to run the income_prediction demo, I found that the script and README.md have references to an older lib. Fixed it in this fork!!

Best,
b.


hectorgon:
Thanks for fixing this!

PR #23, opened 2015-06-11
Add margin to spline trainer

hectorgon:
@deerzq @fenglinlfl @tension7 add margin paramter to the spline model's hinge loss


deerzq:
:+1:

PR #25, opened 2015-06-16
Added missing config parameter linfinity_cap with default value

Zaidos:
While executing `sh job_runner.sh TrainModel` in the income prediction demo, there is an error thrown due to a missing configuration parameter.

This pull request adds the parameter with the default value found in code.


hectorgon:
Thanks for the fix!

PR #26, opened 2015-06-21
Remove dependency on project version and set all demo versions to 1.0…

hectorgon:
…. Add instructions on how to download the adult census data

@deerzq


deerzq:
LGTM

PR #27, opened 2015-06-23
Multiscale splines

hectorgon:
Add support for training multiscale B-splines and also correct a small bias from the dropout.

@deerzq @fenglinlfl @sdemars

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/SplineTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/SplineTrainer.scala
@@ -19,6 +19,7 @@ import org.slf4j.{Logger, LoggerFactory}
 
 import scala.collection.JavaConversions._
 import scala.collection.JavaConverters._
+import scala.collection.mutable.HashMap
 import scala.util.Try
 
 object SplineTrainer {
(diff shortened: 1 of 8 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/SplineTrainer.scala
```diff
@@ -352,7 +510,7 @@ object SplineTrainer {
                   dropout : Double,
                   margin : Double) : Double = {
     val flatFeatures = Util.flattenFeatureWithDropout(fv, dropout)
-    val prediction = model.scoreFlatFeatures(flatFeatures)
+    val prediction = model.scoreFlatFeatures(flatFeatures) / (1.0 - dropout)
```
deerzq: I don't quite get this change. I thought when training with dropout, the updating process should be the same. But at testing time, we need to rescale based on the dropout rate: "At test time, we use the “mean network” that contains all of the hidden units but with their outgoing weights halved to compensate for the fact that twice as many of them are active." so we might need to make the change at the end of the whole training process and scale the weight by weight = weight * (1.0 - dropout rate)

hectorgon: Ok let me revert this tomorrow then since there are additional calibration steps and it doesn't change the performance AUC wise before.

hectorgon: Double checking http://www.cs.toronto.edu/~rsalakhu/papers/srivastava14a.pdf
"In this paper, we described dropout as a method where we retain units with probability p
at training time and scale down the weights by multiplying them by a factor of p at test time.
Another way to achieve the same effect is to scale up the retained activations by multiplying
by 1/p at training time and not modifying the weights at test time."
I'm doing it by the 2nd way so adding a comment

deerzq: :+1:

deerzq:
Adding some comments about the dropout method, then LGTM!

PR #30, opened 2015-07-17
Typo fix

sagivo:


hectorgon:
Thanks for the fix! This was merged in the PR just before yours so I am not double merging it, but thanks so much for finding it!

PR #32, opened 2015-07-31
remove value tuple's key to reduce shuffle size

CanoeFZH:
The value tuple of reduceByKey has part (x._1._1) which is same to the key of reduceByKey. Thus, that part can be removed to reduce shuffle size.


hectorgon:
Good catch, thanks for the optimization

PR #33, opened 2015-08-03
Clean up demo's

janovergoor:
Clean up the two demo's a bit:

* tiny fixes to readme
* resolve some erroneous links between the demos
* removed spurious classes

@hectorgon


hectorgon:
Thanks for the fixes!

PR #34, opened 2015-08-10
add debugScoreComponent for linear model

deerzq:
I was trying to run debugCompare for the mini model, then I realized we don't have this in aerosolve

to: @hectorgon @sdemars

### DIFF
--- a/core/src/test/java/com/airbnb/aerosolve/core/models/LinearModelTest.java
+++ b/core/src/test/java/com/airbnb/aerosolve/core/models/LinearModelTest.java
@@ -1,5 +1,6 @@
 package com.airbnb.aerosolve.core.models;
 
+import com.airbnb.aerosolve.core.DebugScoreRecord;
 import com.airbnb.aerosolve.core.FeatureVector;
 import com.airbnb.aerosolve.core.ModelHeader;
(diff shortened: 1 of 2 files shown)


hectorgon:
LGTM thanks for doing this!

core/src/test/java/com/airbnb/aerosolve/core/models/LinearModelTest.java
```diff
@@ -1,16 +1,19 @@
 package com.airbnb.aerosolve.core.models;
 
+import com.airbnb.aerosolve.core.DebugScoreRecord;
 import com.airbnb.aerosolve.core.FeatureVector;
 import com.airbnb.aerosolve.core.ModelHeader;
 import com.airbnb.aerosolve.core.ModelRecord;
 import com.airbnb.aerosolve.core.util.Util;
 
+import jdk.nashorn.internal.runtime.Debug;
```
hectorgon: I don't think this is supposed to be here

deerzq: you are right, looks like intellij automatically added that, will remove.

PR #35, opened 2015-08-12
update batch score calibrator

deerzq:
Update the ScoreCalibrator.trainBatchMLE 
- change training parameters so that they are consistent with the parameters of trainSGD
- let the input be an array instead of an RDD so that we can use it to train multiple Platt scaling model in parallel
- update the test

to: @hectorgon @fenglinlfl


hectorgon:
LGTM

PR #36, opened 2015-08-13
add null check

deerzq:
- example may contain features that are not in the model, therefore adding the check
- also lower the threshold for BoostedStumpsModelTest 
@hectorgon


hectorgon:
LGTM thanks for fixing this

PR #37, opened 2015-08-17
[LinearRankerTrainer] random sample in each iteration

deerzq:
Apply random sampling in each iteration, so that this is consistent with SplineTrainer. This is because coalesce() does not really reshuffle each time it is called when the input parameters are the same.

@hectorgon


hectorgon:
LGTM

PR #38, opened 2015-08-29
[SplineModel] allow initialization with an existing model

deerzq:
This allows the spline trainer to initialize the model with an existing spline model and do incremental training.

to: @hectorgon @tension7 @sdemars

### DIFF
--- a/training/src/test/scala/com/airbnb/aerosolve/training/SplineTrainerTest.scala
+++ b/training/src/test/scala/com/airbnb/aerosolve/training/SplineTrainerTest.scala
@@ -1,9 +1,13 @@
 package com.airbnb.aerosolve.training
 
 import java.io.{StringReader, BufferedWriter, BufferedReader, StringWriter}
+import java.util
 
-import com.airbnb.aerosolve.core.models.ModelFactory
+import com.airbnb.aerosolve.core.models.SplineModel.WeightSpline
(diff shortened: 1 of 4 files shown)


training/src/test/scala/com/airbnb/aerosolve/training/SplineTrainerTest.scala
```diff
@@ -199,4 +225,42 @@ class SplineTrainerTest {
...
+    // add an existing
+    val weights = model.getWeightSpline.asScala
+    for (familyMap <- weights) {
+      for (featureMap <- familyMap._2.asScala) {
+        val family = familyMap._1
+        val feature = featureMap._1
+        val spline = featureMap._2.spline
+        log.info(("family=%s,feature=%s,minVal=%f, maxVal=%f, weights=%s")
+                   .format(family, feature, spline.getMinVal, spline.getMaxVal, spline.toString))
+        if (family.equals("A")) {
```
hectorgon: I would use a case match but LGTM as is

hectorgon:
LGTM, good catch with the linear save. This only occurred for the calibration right? The trainer AFAIK saved the model directly so was not affected by the bug.

deerzq:
Right, since we haven't done any linear model with calibration, this doesn't affect any of our previous models.

PR #39, opened 2015-08-31
Initial refactor in preparation for ranking loss

hectorgon:
- Moved params into a case class
- Count all the features in all the examples in an example
@deerzq


deerzq:
since I just merged some code in SplineTrainer, please rebase before merge. Otherwise, LGTM

PR #40, opened 2015-09-03
Rank and hinge loss

hectorgon:
Add a new list based ranking and hinge loss

@deerzq @tension7 @sdemars

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerUtils.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerUtils.scala
@@ -181,14 +181,4 @@ object LinearRankerUtils {
       CompressedExample(getFeatures(ex), Array[(String, String)](), label)
     })
   }
-  
-  def getLabel(fv : FeatureVector, rankKey : String, threshold : Double) : Double = {
-    val rank = fv.floatFeatures.get(rankKey).asScala.head._2
-    val label = if (rank <= threshold) {
-      -1.0
-    } else {
-      1.0
-    }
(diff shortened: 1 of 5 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerUtils.scala
```diff
@@ -168,4 +181,14 @@ object LinearRankerUtils {
...
+  def getLabel(fv : FeatureVector, rankKey : String, threshold : Double) : Double = {
+    val rank = fv.floatFeatures.get(rankKey).asScala.head._2
+    val label = if (rank <= threshold) {
+      -1.0
+    } else {
+      1.0
+    }
+    return label
+  }
 }
```
deerzq: These seem to be something that can be used for all models, why don't we move this to TrainingUtils

training/src/test/scala/com/airbnb/aerosolve/training/SplineRankingTrainerTest.scala
```diff
@@ -0,0 +1,139 @@
...
+    for (i <- 0 to 10) {
+      addItem(example, likes_mod, i)
+    }
+    example
+  }
+  def addItem(example : Example, likes_mod : Int, i : Int) = {
+    val rank : Double = if (i % likes_mod == 0) {
+      1.0
+    } else {
+      0.0
```
deerzq: -1.0? since your rank_threshold = 0.0

hectorgon:
Thank you fixed both issues

PR #42, opened 2015-09-11
Random forest implementation

hectorgon:
Simple random forest implementation

@deerzq @tension7 @sdemars


deerzq:
The code LGTM, but this is not strictly following the definition of random forest. It might be less confusing if we rename this as something like DecisionTreeBagger?

hectorgon:
Ok I will just rename it to Forest model so that I can do boosting later and not be restricted to bagging

PR #43, opened 2015-09-16
Add in gini and information gain split criteria

hectorgon:
@deerzq @tension7

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
@@ -172,7 +172,7 @@ object DecisionTreeTrainer {
             case "gini" => {
               // Using negative gini since we are maximizing.
               val gini = - (f1 * (p1 * (1.0 - p1)  + n1 * (1.0 - n1)) +
-                  f2 * (n2 * (1.0 - n2) + p2 * (1 - p2)))
+                  f2 * (n2 * (1.0 - n2) + p2 * (1.0 - p2)))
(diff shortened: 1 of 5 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
```diff
@@ -156,14 +162,53 @@ object DecisionTreeTrainer {
...
+          val n1 = rightNeg / rightCount
+          val f1 = rightCount / (leftCount + rightCount)
+          val p2 = leftPos / leftCount
+          val n2 = leftNeg / leftCount
+          val f2 = leftCount / (leftCount + rightCount)
+          splitCriteria match {
+            case "gini" => {
+              // Using negative gini since we are maximizing.
+              val gini = - (f1 * (p1 * (1.0 - p1)  + n1 * (1.0 - n1)) +
+                  f2 * (n2 * (1.0 - n2) + p2 * (1 - p2)))
```
spencerde: super nitpick `1.0 - p2` for consistency

spencerde:
Nice LGTM.  

The unit tests don't actually differentiate between the different split methods it just tests that those strings are accepted.  If there was a clever way to easily get to a different fraction correct w/ the different split criteria that would be even better but it's probably fine.

hectorgon:
The unit tests just check the outcome yeah doesn't actually test the formula. The fractionCorrect sort of varies from run to run so I use a conservative lower bound.

spencerde:
Checking the outcome is still fairly good protection I see what you are saying :+1:

PR #46, opened 2015-09-17
Sampling strategy for boosting

hectorgon:
Add a first sampling strategy as uniform sampling is really slow
@deerzq


deerzq:
LGTM!

PR #47, opened 2015-09-17
Faster version of take first

hectorgon:
For some reason the take on a partition takes a very long time .. just make a buffer and skip the test.
@deerzq


deerzq:
LGTM

PR #49, opened 2015-09-18
fix the evaluation

deerzq:
Assume label = 0 corresponds to negative label because it seems people also tend to use 0 as negative label. Without the fix, we are not counting true negatives for label=0
@hectorgon


hectorgon:
LGTM

buptyc:
LGTM

PR #50, opened 2015-09-22
Add rmse evaluation for regression

spencerde:
For regression the binary classification metrics don't really make sense anyway and don't calculate RMSE in the same way.  This adds a new short evaluation for regression which for now is just train and hold RMSE.  (Let me know if naming seems okay or should be changed as some of those other metrics might make sense for regression on binary events, this is intended for regression where the labels are real values so could make name more specific if you think that makes sense).

to: @hectorgon @deerzq


hectorgon:
LGTM

PR #52, opened 2015-10-27
Support regression trees in DecisionTreeTrainer

yolken:
- Modify existing DecisionTreeTrainer to support both classification- and regression-type splits; see http://www.stat.cmu.edu/~cshalizi/350-2006/lecture-10.pdf for a good overview of the latter.
- Misc. cleanups to associated files

@hectorgon can you take a look? Thanks.

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
@@ -149,6 +149,7 @@ object DecisionTreeTrainer {
           sum += labelValue
         }
 
+        // In regression case, leaf is the average of all the associated values
         rec.setFeatureWeight(sum / count)
       }
       case _ => {
--- a/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
@@ -51,7 +51,6 @@ object DecisionTreeTrainer {
         .map(x => Util.flattenFeature(x.example(0)))
         .filter(x => x.contains(rankKey))
         .take(candidateSize)
-        .toArray
     
     val stumps = new util.ArrayList[ModelRecord]()
     stumps.append(new ModelRecord)
@@ -59,8 +58,6 @@ object DecisionTreeTrainer {
     
     val model = new DecisionTreeModel()
     model.setStumps(stumps)
-
-    log.info("Dot model=> %s".format(model.toDot))
 
     model
   }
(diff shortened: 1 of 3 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
```diff
@@ -129,94 +167,172 @@ object DecisionTreeTrainer {
...
+    for (example <- examples) {
+      val response = BoostedStumpsModel.getStumpResponse(candidateOpt.get, example)
+      val labelValue = example.get(rankKey).asScala.head._2
+
+      // Using "online algorithm" for computing mean and sum-squared errors as
+      // described in https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
+      if (response) {
+        rightCount += 1
+        val delta = labelValue - rightMean
+        rightMean += delta / rightCount
```
hectorgon: This is not the right way to compute the variance .. you should accumulate the sum and sum squared and only divide by the count at the very end outside of the loop

training/src/test/scala/com/airbnb/aerosolve/training/TrainingTestHelper.scala
```diff
@@ -47,4 +44,22 @@ object TrainingTestHelper {
...
+  def makeRegressionExamples = {
+    val examples = ArrayBuffer[Example]()
+    val label = ArrayBuffer[Double]()
+    val rnd = new java.util.Random(1234)
+
+    for (i <- 0 until 200) {
+      val x = 4.0 * (2.0 * rnd.nextDouble() - 1.0)
+      val y = 4.0 * (2.0 * rnd.nextDouble() - 1.0)
+
+      val quadradtic = x * x - 2 * y * y - 0.5 * x + 0.2 * y
```
hectorgon: Can we include a small flat region by adding max(1.0, quadratic) to see if the decision tree compresses that region with a small number of splits?

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
hectorgon: Clarified that this is the online variant which is more numerically stable (Welford)

hectorgon:
LGTM

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
yolken: Made the comment more precise and also made the associated unit test a little more complex to verify that it works correctly.

training/src/test/scala/com/airbnb/aerosolve/training/TrainingTestHelper.scala
yolken: Changed the curve to have some flat regions and added a test that the algorithm does the right thing in the associated regions.

yolken:
@hectorgon addressed your feedback. Feel free to take another look.

PR #55, opened 2015-11-04
Add regression in spline trainer

deerzq:
Add regression for spline trainer
@hectorgon


training/src/main/scala/com/airbnb/aerosolve/training/SplineTrainer.scala
```diff
@@ -605,6 +615,21 @@ object SplineTrainer {
     return loss
   }
 
+  def updateRegressor(model: SplineModel,
+                      fv: FeatureVector,
+                      label: Double,
+                      params : SplineTrainerParams) : Double = {
+    val flatFeatures = Util.flattenFeatureWithDropout(fv, params.dropout)
+    val prediction = model.scoreFlatFeatures(flatFeatures) / (1.0 - params.dropout)
+    val loss = math.abs(prediction - label) // absolute difference
```
hectorgon: Isn't it max(0, abs(p-l) - epsilon) ? But this is just reporting the loss so no big deal whichever you think seems better.

hectorgon:
LGTM

PR #58, opened 2015-11-16
Add Percentage Error to Regression EVAL

jq:
@deerzq @hectorgon 
Percentage will help eval relative error of regression


training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
```diff
@@ -232,7 +236,12 @@ object Evaluation {
     val out = collection.mutable.ArrayBuffer[(String, Double)]()
 
     val prefix = if (record.is_training) "TRAIN_" else "HOLD_"
-    out.append((prefix + "SQERR", (record.label - record.score) * (record.label - record.score)))
+    val diff = (record.label - record.score)
+    val sqErr = diff * diff;
+    val diffPercentage = diff / record.score
```
hectorgon: possible divide by zero here

deerzq: this only makes sense for regression and it should be diff/record.label, when record.label is zero, you can add a smoothing constant term (diff + c) / (record.label + c)

jq: Thanks for review!! fixed and uploaded.

PR #59, opened 2015-11-18
Add in squared loss for regression in linear model

hectorgon:
Squared loss regression
@deerzq @yolken @spencerde

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerTrainer.scala
@@ -75,6 +75,7 @@ object LinearRankerTrainer {
     loss match {
       case "ranking" => rankingTrain(sc, input, config, key, numBags, weights, iteration)
       case "regression" => regressionTrain(sc, input, config, key, numBags, weights, iteration)
+      case "regression2" => regression2Train(sc, input, config, key, numBags, weights, iteration)
       case "hinge" => classificationTrain(sc, input, config, key, numBags, weights, iteration)
       case "logistic" => logisticTrain(sc, input, config, key, numBags, weights, iteration)
       case _ => {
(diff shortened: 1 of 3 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerTrainer.scala
```diff
@@ -75,6 +75,7 @@ object LinearRankerTrainer {
     loss match {
       case "ranking" => rankingTrain(sc, input, config, key, numBags, weights, iteration)
       case "regression" => regressionTrain(sc, input, config, key, numBags, weights, iteration)
+      case "regression2" => regression2Train(sc, input, config, key, numBags, weights, iteration)
```
deerzq: `regression2` sounds not so informative

training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerTrainer.scala
```diff
@@ -149,6 +150,66 @@ object LinearRankerTrainer {
...
+                                        sum = newGradSum)
+            weightMap.put(v, (newWeight, newGradSum))
+          })
+          weightMap.put(lossKey, (lossEntry._1 + loss, lossEntry._2 + 1.0))          
+        })
+      })
+      weightMap
+        .iterator
+    })
+  }
```
deerzq: can we ask the user to pass in regression loss function name in config, then modify regressionTrain to take different updates based on the loss function name?

training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerTrainer.scala
hectorgon: Fixed

training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerTrainer.scala
hectorgon: It's not that much code and then we'd have problems with epsilon which we don't require for L2 so I'd rather keep it as a separate function instead of having case statements inside the loop

PR #60, opened 2015-11-19
additive model

deerzq:
- Implement AdditiveModel which takes more flexible forms of functions. In addition to Spline, it also takes linear functions. 
- Add an abstract function class, Spline and Linear all inherit this class.
- Add unit test, fix a small bug in LinearModelTest.java 
- implement additive model trainer

@hectorgon


core/src/main/java/com/airbnb/aerosolve/core/util/AbstractFunction.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.Setter;
+
+import java.io.Serializable;
+
+/**
+ * Base class for functions
+ */
+abstract public class AbstractFunction implements Serializable{
+  @Getter @Setter
+  protected String functionForm; // default function is Spline
```
hectorgon: There are millions of these maybe make it an enum or something smaller?

hectorgon:
LG except for the spline function as a string ... can we make it an enum?

deerzq:
@hectorgon thanks for the comments. FunctionForm is changed to enum now and also added an additive model trainer, PTAL

hectorgon:
I don't see the min and max being used in the linear function, don't we need to normalize between features of very different dynamic range? Should it not compute a z = (x - minVal) / (maxVal - minVal) and compute the linear term over z?

deerzq:
@hectorgon Good point, I was thinking doing the normalization before feeding the feature into the trainer. But I can move this inside the model training part.

deerzq:
@hectorgon I just added minVal and maxVal in linear function and normalize input feature inside the linear function.

hectorgon:
LGTM

PR #61, opened 2015-11-23
Make a simple linearly separable unit test for boosted stumps model

hectorgon:
Since stumps without crosses are unable to separate a quadratic make a linear decision surface for the unit test.

@deerzq @yolken


yolken:
LGTM

PR #63, opened 2015-12-01
add multiscale spline train in additive model

deerzq:
to: @hectorgon @yolken @jq


hectorgon:
Actually I was thinking if you already have additive models won't a spline + linear additive function work better than multi-scale splines?

hectorgon:
:+1:

deerzq:
I guess multiscale spline + linear might work better than spline + linear

deerzq:
@jq nothing related to that test was changed, the test itself may fail ocassionally

PR #64, opened 2015-12-04
[Pricing] add CustomMultiscaleQuantizeTransform

deerzq:
Add a transform that allows to select or exclude specific features for quantization
@hectorgon


hectorgon:
:+1:

PR #65, opened 2015-12-08
try to speed up the trainer by shuffling only once at the beginning

deerzq:
- shuffle the transformed examples once at the beginning
- then persist it at disk level
- no shuffle in subsequence coalesce
this change works the best when subsample rate is smaller than the 1 / (number of iterations), otherwise the same data may end up in the same partition for different iterations. But hopefully the problem won't be so severe 

- try this for additive model trainer first, if it works will update for other trainers.
@hectorgon


hectorgon:
:+1:

PR #66, opened 2015-12-10
modify substract and divide transform

deerzq:
- make the behavior of divide and subtract transforms more consistent
- allow user to specify features to be included in the transform
- change the output feature name of subtract transform to be different from the original key

@hectorgon @askeys


hectorgon:
:+1: nice fix!

PR #67, opened 2015-12-10
Kernel machines

hectorgon:
Kernel machines supporting classification and regression

@deerzq @yolken @spencerde @askeys

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/AbstractModel.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/AbstractModel.java
@@ -5,10 +5,12 @@ import java.io.BufferedWriter;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.List;
+import java.util.Map;
 
 import com.airbnb.aerosolve.core.DebugScoreRecord;
 import com.airbnb.aerosolve.core.FeatureVector;
 import com.airbnb.aerosolve.core.ModelHeader;
+
 import lombok.Getter;
 import lombok.Setter;
 /**
@@ -45,6 +47,10 @@ public abstract class AbstractModel implements Model, Serializable {
   // returns probability: 1 / (1 + exp(-(offset + scale * score))
   public double scoreProbability(float score) {
     return 1.0 / (1.0 + Math.exp(-(offset + slope * score)));
+  }
+  
+  // Optional method implemented by online updatable models e.g. Spline, RBF
+  public void onlineUpdate(float grad, float learningRate, Map<String, Map<String, Double>> flatFeatures) {
   }
 
   // Helper function for FOBOS updates.
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/AbstractModel.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/AbstractModel.java
@@ -51,6 +51,7 @@ public abstract class AbstractModel implements Model, Serializable {
   
   // Optional method implemented by online updatable models e.g. Spline, RBF
   public void onlineUpdate(float grad, float learningRate, Map<String, Map<String, Double>> flatFeatures) {
+    assert(false);
   }
 
   // Helper function for FOBOS updates.
--- a/core/src/main/java/com/airbnb/aerosolve/core/util/StringDictionary.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/util/StringDictionary.java
@@ -1,6 +1,8 @@
 package com.airbnb.aerosolve.core.util;
 
 import com.airbnb.aerosolve.core.FeatureVector;
+import com.airbnb.aerosolve.core.DictionaryEntry;
+import com.airbnb.aerosolve.core.DictionaryRecord;
 
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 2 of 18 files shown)


core/src/main/java/com/airbnb/aerosolve/core/util/StringDictionary.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+  
+  public FloatVector makeVectorFromSparseFloats(Map<String, Map<String, Double>> sparseFloats) {
+    FloatVector vec = new FloatVector(dictionary.getEntryCount());
+    for (Map.Entry<String, Map<String, Double>> kv : sparseFloats.entrySet()) {
+      for (Map.Entry<String, Double> feat : kv.getValue().entrySet()) {
+        DictionaryEntry entry = getEntry(kv.getKey(), feat.getKey());
+        if (entry != null) {
+          vec.values[entry.index] = (float) entry.scale * (feat.getValue().floatValue() - (float) entry.mean);
+        }
```
deerzq: If I remember correctly, we also put string features in `sparseFloats` and the value would be 1.0? then what is the `scale` and `mean` for them

core/src/main/java/com/airbnb/aerosolve/core/util/SupportVector.java
```diff
@@ -0,0 +1,90 @@
...
+import lombok.Setter;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Random;
+
+import com.airbnb.aerosolve.core.FunctionForm;
+import com.airbnb.aerosolve.core.ModelRecord;
+
+public class SupportVector implements Serializable {
```
deerzq: It seems `SupportVector` can also extend `AbstractFuction`, then RBF sv and  ARC_COSINE sv can be subclasses

core/src/main/java/com/airbnb/aerosolve/core/util/StringDictionary.java
```diff
@@ -0,0 +1,76 @@
...
+import lombok.Setter;
+
+import java.io.Serializable;
+import java.util.*;
+import java.util.AbstractMap.SimpleEntry;
+import java.util.Map.Entry;
+
+/**
+ * A class that maps strings to indices.
+ */
```
deerzq: perhaps add more details about this class, since it actually deals with both string features and float features

training/src/main/scala/com/airbnb/aerosolve/training/KernelTrainer.scala
```diff
@@ -0,0 +1,126 @@
...
+    val stats = TrainingUtils.getFeatureStatistics(minCount, examples)
+    log.info(s"Dictionary size is ${stats.size}")
+    
+    for (stat <- stats) {
+      val (family, feature) = stat._1
+      if (family != rankKey) {
+        val mean = stat._2.mean
+        val variance = Math.max(1e-6, stat._2.variance)
+        val scale = Math.sqrt(1.0 / variance)
+        dictionary.possiblyAdd(family, feature, mean, scale)
```
deerzq: for stringFeatures, variance would be very small then scale would be super large, and mean would be just 1.0, then the corresponding value in the vector generated by the dictionary would be 0, we may want to make it 1.0 instead?

training/src/main/scala/com/airbnb/aerosolve/training/KernelTrainer.scala
```diff
@@ -0,0 +1,126 @@
...
+        
+    val candidates = examples.take(candidateSize)
+
+    // Super simple SGD trainer. Mostly to get the unit test to pass
+    for (candidate <- candidates) {
+      val gradient = computeGradient(model, candidate.example(0), loss, rankKey, rankThreshold)
+      if (gradient != 0.0) {
+        val flatFeatures = Util.flattenFeature(candidate.example(0));
+        model.onlineUpdate(gradient, learningRate, flatFeatures)
+      }
```
deerzq: seems we are fixing all the support vectors the same all the time during the training, perhaps we can delete some support vectors if their responses are very small.

core/src/main/java/com/airbnb/aerosolve/core/models/AbstractModel.java
```diff
@@ -46,6 +48,10 @@ abstract public float debugScoreItem(FeatureVector combinedItem,
   public double scoreProbability(float score) {
     return 1.0 / (1.0 + Math.exp(-(offset + slope * score)));
   }
+  
+  // Optional method implemented by online updatable models e.g. Spline, RBF
+  public void onlineUpdate(float grad, float learningRate, Map<String, Map<String, Double>> flatFeatures) {
```
yolken: Throw an exception or log an error here so that people know they're calling something with no effect?

core/src/main/java/com/airbnb/aerosolve/core/util/SupportVector.java
hectorgon: I think abstract function was made too specific, with things like Linfinity cap, updating at X ... support vector takes in vectors too not scalars so I don't think I will do this change since it is a vector function that returns a scalar instead of a scalar function taking a scalar.

core/src/main/java/com/airbnb/aerosolve/core/util/StringDictionary.java
hectorgon: Ok done

training/src/main/scala/com/airbnb/aerosolve/training/KernelTrainer.scala
hectorgon: Ok done also remember missing string features are 0 in the dense case so the variance is actually p * (1-p) where p is the probability of getting the string feature on. The variance is only 0 for the bias term as it is always on.

training/src/main/scala/com/airbnb/aerosolve/training/KernelTrainer.scala
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+import org.slf4j.{Logger, LoggerFactory}
+
+import scala.util.Random
+import scala.util.Try
+import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
+
+// The decision tree is meant to be a prior for the spline model / linear model
```
yolken: Comment for the wrong class?

core/src/main/java/com/airbnb/aerosolve/core/models/AbstractModel.java
hectorgon: Done

training/src/main/scala/com/airbnb/aerosolve/training/KernelTrainer.scala
hectorgon: Oops fixed

training/src/main/scala/com/airbnb/aerosolve/training/KernelTrainer.scala
hectorgon: I'll add that if kernel models are used offline .. right now the trainer is mostly for the unit test ... the main use case is online learning.

hectorgon:
@deerzq @yolken Please take another look. The trainer is very rudimentary and I didn't put much effort in it, mostly for the unit tests but need to make sure that the interface is good enough to function in the online learning context.

core/src/main/java/com/airbnb/aerosolve/core/util/StringDictionary.java
hectorgon: if prob (x = 1) = p where X is the string feature,
then mean = p
scale = 1 / sqrt(p * (1-p))

training/src/main/scala/com/airbnb/aerosolve/training/KernelTrainer.scala
hectorgon: Added it as a TODO up top if/when I implement regularization

core/src/main/java/com/airbnb/aerosolve/core/util/SupportVector.java
deerzq: cool

training/src/main/scala/com/airbnb/aerosolve/training/KernelTrainer.scala
```diff
@@ -0,0 +1,137 @@
...
+    val stats = TrainingUtils.getFeatureStatistics(minCount, examples)
+    log.info(s"Dictionary size is ${stats.size}")
+    
+    for (stat <- stats) {
+      val (family, feature) = stat._1
+      if (family != rankKey) {
+        val mean = stat._2.mean
+        val variance = if (stat._2.variance < 1e-6) 1.0 else stat._2.variance
+        var scale = Math.sqrt(1.0 / variance)
+        dictionary.possiblyAdd(family, feature, mean, scale)
```
deerzq: Not sure if I am missing anything. But for a string feature, if it's missing, featureStatistics won't see it, therefore the `mean` here would be 1, then inside `makeVectorFromSparseFloats`, the z-transform `(float) entry.scale * (feat.getValue().floatValue() - (float) entry.mean);` would transform the value of the String feature to 0.0, since for String feature its value = mean. And then when we see an example who doesn't contain this String feature, by default its value would also be 0.0, since the `FloatVector` constructor would initialize a all-zero vector. 

Then the question is how do we discriminate a missing String feature and a String feature that exists?

hectorgon:
@deerzq good catch, please take another look, fixed the string feature issue

deerzq:
LGTM!

PR #68, opened 2015-12-10
revert the repartition change

deerzq:
- this seems bring in some disk full problem for our cluster, so revert it.

@hectorgon


hectorgon:
:+1:

PR #69, opened 2015-12-14
Reinforcement learning utiliites

hectorgon:
@yolken @deerzq @askeys

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/util/ReinforcementLearning.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/util/ReinforcementLearning.java
@@ -1,0 +1,59 @@
+package com.airbnb.aerosolve.core.util;
+
+/**
+ * Utilities for reinforcement learning
+ */
+
+import com.airbnb.aerosolve.core.models.AbstractModel;
+import com.airbnb.aerosolve.core.FeatureVector;
+import com.google.common.hash.HashCode;
+import com.google.common.hash.Hasher;
+import com.google.common.hash.Hashing;
+
+import org.apache.commons.codec.binary.Base64;
+import org.apache.thrift.TDeserializer;
+import org.apache.thrift.TSerializer;
+import org.apache.thrift.TBase;
+
+import java.io.Serializable;
+import java.util.*;
+
+public class ReinforcementLearning implements Serializable {
+  // Updates a model using SARSA
+  // https://en.wikipedia.org/wiki/State-Action-Reward-State-Action
+  // If the nextState is terminal, pass in null
+  public static void updateSARSA(AbstractModel model,
+                                 FeatureVector stateAction,
+                                 float reward,
+                                 FeatureVector nextStateAction,
+                                 float learningRate,
+                                 float decay) {
+    Map<String, Map<String, Double>> flatSA = Util.flattenFeature(stateAction);
+    float nextQ = 0.0f;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/airbnb/aerosolve/core/util/ReinforcementLearning.java
```diff
@@ -0,0 +1,60 @@
...
+public class ReinforcementLearning implements Serializable {
+  // Updates a model using SARSA
+  // https://en.wikipedia.org/wiki/State-Action-Reward-State-Action
+  // If the nextState is terminal, pass in null
+  public static void updateSARSA(AbstractModel model,
+                                 FeatureVector stateAction,
+                                 float reward,
+                                 FeatureVector nextStateAction,
+                                 float learningRate,
+                                 float decay) {
```
yolken: I would call this "discountRate" or "discountFactor" to match the terminology in the literature.

core/src/main/java/com/airbnb/aerosolve/core/util/ReinforcementLearning.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+   
+    float currentQ = model.scoreItem(stateAction);
+    float expectedQ = reward + decay * nextQ;
+    float grad = currentQ - expectedQ;
+    model.onlineUpdate(grad, learningRate, flatSA);
+  }
+
+  // Picks a random action with probability epsilon.
+  public static int epsilonGreedyPolicy(AbstractModel model, ArrayList<FeatureVector> stateAction, float epsilon, Random rnd) {
```
yolken: Implement softmax as well? That might be better than epsilon greedy in some situations.

core/src/main/java/com/airbnb/aerosolve/core/util/ReinforcementLearning.java
hectorgon: Done

core/src/main/java/com/airbnb/aerosolve/core/util/ReinforcementLearning.java
hectorgon: Done

core/src/test/java/com/airbnb/aerosolve/core/util/ReinforcementLearningTest.java
```diff
@@ -0,0 +1,181 @@
...
+    float prevReward = 0.0f;
+    FeatureVector currStateAction = null;
+    float epsilon = 0.1f;
+    float learningRate = 0.1f;
+    if (epoch > 90) {
+      epsilon = 0.0f;
+      learningRate = 0.001f;
+    }
+    boolean done = false;
+    while (!done) {
```
yolken: I think this works, but it might be a little easier to follow if a full SARSA update is done at every iteration (as opposed to deferring it one period).

The revised approach would be:
1. Initialize currStateAction with x=0, y=0, pick initial action
2. For each iteration:
  1. Evaluate reward, next (x, y) based on currStateAction
  2. If (x, y) is not terminal, pick greedy action for next x, next y and set in nextStateAction; otherwise, nextStateAction is null
  3. Do full SARSA update using (currStateAction, reward, nextStateAction)
  4. Set currStateAction to nextStateAction

yolken:
LGTM

PR #72, opened 2015-12-19
Date Related Transform

seckcoder:
Add date related transform including:

1. Date Diff Transform: difference in days between two dates
2. Year of Date Transform: get year from date
3. Month of Date Transform: get month from date
4. Day of Year Transform: get day of year from date
5. Day of Month Transform: get day of month from date
6. Day of Week Transform: get day of week from date


jq:
can you share more code between the transformer?

core/src/main/thrift/MLSchema.thrift
```diff
@@ -42,6 +42,18 @@ struct Example {
...
+  2: optional double mean;
+  3: optional double scale; 
+}
+
+struct DictionaryRecord {
+  1: optional map<string, map<string, DictionaryEntry>> dictionary;
+  2: optional i32 entryCount
+}
+
+
```
deerzq: why moving these two up?

seckcoder: Using a type before it's defined is only supported by recent versions of thrift compiler. As far as I know, if you use thrift version 0.8.0 to compile the original MLSchema.thrift file, an error "Type DictionaryRecord has not been defined" will be reported. See this [link](http://webcache.googleusercontent.com/search?q=cache:2OQpHXHwiQUJ:https://mail-archives.apache.org/mod_mbox/thrift-user/201001.mbox/%253C977a11431001150435m5b2f9536l4f835cc0d40d86fd%40mail.gmail.com%253E+&cd=1&hl=en&ct=clnk&gl=us) for more details.

core/src/main/java/com/airbnb/aerosolve/core/transforms/DayOfMonthTransform.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Calendar;
+
+/**
+ * Get day of month from date
+ */
+public class DayOfMonthTransform extends DateTransform {
+  @Override
+  public Double doDateTransform(Calendar cal) {
+    // First day of month is 1
+    return (double)cal.get(Calendar.DAY_OF_MONTH);
```
jq: instead of creating new class, how about a config, field2,  = day_of_month,  since in  DateDiffTransform , its store date, maybe we just need one class?

seckcoder: I think we can put "day of week", "day of month", "day of year", "month of date" and "year of date" into one class. But it might be a little confusing of mixing the code of "date diff" with the other five "functions" since they have different functionality. What do you think? @jq

jq: agree

core/src/main/java/com/airbnb/aerosolve/core/transforms/DateDiffTransform.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class DateDiffTransform extends Transform {
+  private String fieldName1;
+  private String fieldName2;
+  private String outputName;
+  private final static SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd");
+
+  @Override
+  public void configure(Config config, String key) {
+    fieldName1 = config.getString(key + ".field1");
```
jq: how about having a BaseTransfrom, to take care boilerplate code, i.e. get config, maybe we can set how many fields, then to does the config?

core/src/main/java/com/airbnb/aerosolve/core/transforms/DateDiffTransform.java
```diff
@@ -0,0 +1,67 @@
...
+      return ;
+    }
+
+    Set<String> feature1 = stringFeatures.get(fieldName1);
+    Set<String> feature2 = stringFeatures.get(fieldName2);
+    if (feature1 == null || feature2 == null) {
+      return ;
+    }
+
+    Map<String, Double> output = new HashMap<>();
```
jq: boilerplate code here may be shared in Base too?

core/src/test/java/com/airbnb/aerosolve/core/transforms/DateDiffTransformTest.java
```diff
@@ -0,0 +1,62 @@
...
+            "}";
+  }
+
+  public FeatureVector makeFeatureVector() {
+    Map<String, Set<String>> stringFeatures = new HashMap<>();
+    Map<String, Map<String, Double>> floatFeatures = new HashMap<>();
+    Set endDates = new HashSet<String>();
+    Set startDates = new HashSet<String>();
+    endDates.add("2009-03-01");
+    startDates.add("2009-02-27");
```
jq: you may need multi years test example

core/src/main/java/com/airbnb/aerosolve/core/transforms/DateDiffTransform.java
seckcoder: I don't quite get the BaseTransform stuff. Do you mean to share the code between all transforms or just DateDiffTransform and DateTransform?

deerzq: @jq I feel the BaseTransform stuffs can be done in a different PR. For this PR, it should be ok do it the current way

jq: @deerzq @seckcoder agreed

deerzq:
LGTM

PR #74, opened 2016-01-08
add a generic math transform

deerzq:
add a point-wise math transform to allow transforming some of our features to log space more easily

to: @hectorgon @yolken @jq

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: java
 os : osx
+jdk: oraclejdk8
 before_install:
   - brew update
   - brew install thrift
   - brew install gradle
-  - "wget --no-cookies --no-check-certificate --header \"Cookie: gpw_e24=http%3A%2F%2Fwww.oracle.com%2F; oraclelicense=accept-securebackup-cookie\" http://download.oracle.com/otn-pub/java/jdk/8u66-b17/jdk-8u66-macosx-x64.dmg"
-  - hdiutil mount jdk-8u66-macosx-x64.dmg
-  - sudo installer -pkg /Volumes/JDK\ 8\ Update\ 66/JDK\ 8\ Update\ 66.pkg -target LocalSystem
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: java
-jdk:
-  - oraclejdk8
 os : osx
 before_install:
   - brew update
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: java
 os : osx
-
+jdk:
+  - oraclejdk8
 before_install:
   - brew update
   - brew install thrift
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatFeatureMathTransform.java
```diff
@@ -0,0 +1,83 @@
...
+    if (floatFeatures == null) {
+      return;
+    }
+    Map<String, Double> feature1 = floatFeatures.get(fieldName1);
+
+    if (feature1 == null) {
+      return;
+    }
+    try {
+      DoubleFunction<Double> func = getFunction();
```
hectorgon: Can we put this in the configure so that we do not get function for every feature?

core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatFeatureMathTransform.java
```diff
@@ -0,0 +1,83 @@
...
+  private DoubleFunction<Double> getFunction() {
+
+    switch (functionName) {
+      case "sin":
+        return (double x) -> Math.sin(x);
+      case "cos":
+        return (double x) -> Math.cos(x);
+      case "log10":
+        // return the original value if x <= 0
+        return (double x) -> x > 0 ? Math.log10(x) : x;
```
hectorgon: Wait so that means x = 0.1 and x = -1 are the same value?
Also I suspect you might want to use log1p as well ( log(x + 1) )

deerzq: yes, but for x < -1, there will be problem, perhaps just throw an error if x < 0?

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: java
 os : osx
```
hectorgon: Actually don't need osx I did it in order to install thrift .. another thing is to just use linux and somehow install thrift

deerzq: we also used brew for installing gradle, it is working well with java8 now after doing brew install, so perhaps we can just keep the osx here?

core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatFeatureMathTransform.java
```diff
@@ -0,0 +1,87 @@
...
+    if (feature1 == null) {
+      return;
+    }
+    Map<String, Double> output = new HashMap<>();
+    for (String key : keys) {
+      Double v = feature1.get(key);
+      if (v != null) {
+        Double result = func.get().apply(v);
+        if (!result.isNaN() && !result.isInfinite()) {
+          output.put(key, result);
```
deerzq: @hectorgon here is the fix, compile passed, will merge

PR #76, opened 2016-01-12
Tentative forest RL algorithm

hectorgon:
WIP has instabilities
@yolken


hectorgon:
Closing this as we realize that decision trees result in flat Q surfaces with no gradient to walk up or down

PR #77, opened 2016-01-13
allow the capped feature to output to a new feature family

deerzq:
@hectorgon as we discussed yesterday, sometimes we may want to keep both capped an non-capped features, therefore add this change


jq:
LGTM

hectorgon:
:+1:

PR #78, opened 2016-01-13
fix possible output overwrite problem for transforms

deerzq:
If a transform outputs to an existing feature family. Some of our transforms would overwrite the content of existing feature family. This PR makes sure all transforms won't have this behavior by always using `Util.getOrCreateFloatFeature` or `Util.getOrCreateStringFeature` to generate new output.

to: @hectorgon @jq @yolken

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/transforms/ApproximatePercentileTransform.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/transforms/ApproximatePercentileTransform.java
@@ -1,6 +1,8 @@
 package com.airbnb.aerosolve.core.transforms;
 
 import com.airbnb.aerosolve.core.FeatureVector;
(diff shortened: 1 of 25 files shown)


jq:
@hectorgon how about add a "protection" field in conf, and once it set, override this feature or this feature override other feature will output error log. 
default is no protection.
@seckcoder  can help this.

hectorgon:
I like this change but just we should check nothing depends on the previous behavior otherwise LGTM

deerzq:
@hectorgon I just checked our code by searching in config using keyword 'output', it seems all previous transforms with 'output' field are either output to new families or have already handled this issue in transform. Not sure about other teams' use case, but I don't think there is any reason to overwrite an entire feature family, if we want to do that we usually just use the delete transform.

PR #79, opened 2016-01-14
fix the concurrent modification error when outputName=fieldName1

deerzq:
I got a concurrent modification error for the case that `outputName = fieldName1`. 
This is a fix.

to: @hectorgon @jq


hectorgon:
:+1:

PR #81, opened 2016-01-22
fix the divide and subtract transform

deerzq:
Still getting the concurrent modification exception, I guess we should always let the user specify the specific keys for feature transform

@hectorgon


hectorgon:
:+1:

PR #82, opened 2016-01-22
Full rank linear model

hectorgon:
A simple full rank linear model for multi-class classification
@deerzq @yolken @christhetree

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/FullRankLinearModel.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/FullRankLinearModel.java
@@ -1,0 +1,124 @@
+package com.airbnb.aerosolve.core.models;
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.*;
+
+import com.airbnb.aerosolve.core.DebugScoreRecord;
+import com.airbnb.aerosolve.core.FeatureVector;
+import com.airbnb.aerosolve.core.LabelDictionaryEntry;
+import com.airbnb.aerosolve.core.ModelHeader;
+import com.airbnb.aerosolve.core.ModelRecord;
+import com.airbnb.aerosolve.core.MulticlassScoringOptions;
+import com.airbnb.aerosolve.core.MulticlassScoringResult;
+import com.airbnb.aerosolve.core.util.Util;
+import com.airbnb.aerosolve.core.util.FloatVector;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+// A full rank linear model that supports multi-class classificaiton.
+// The class vector Y = W' X where X is the feature vector.
+public class FullRankLinearModel extends AbstractModel {
+
+  private static final long serialVersionUID = -849900702679383420L;
+
+  @Getter @Setter
+  private Map<String, Map<String, FloatVector>> weightVector;
+  
+  @Getter @Setter
+  private ArrayList<LabelDictionaryEntry> labelDictionary;
+
+  public FullRankLinearModel() {
+  }
+
+  // In the binary case this is just the score for class 0.
+  @Override
+  public float scoreItem(FeatureVector combinedItem) {
+    Map<String, Map<String, Double>> flatFeatures = Util.flattenFeature(combinedItem);
+    return 0.0f;
+  }
+
+  @Override
+  public float debugScoreItem(FeatureVector combinedItem,
+                              StringBuilder builder) {
+    // TODO(hector_yee) : implement debug.
+    return scoreItem(combinedItem);
+  }
+
+  @Override
+  public List<DebugScoreRecord> debugScoreComponents(FeatureVector combinedItem) {
+    // (TODO) implement debugScoreComponents
+    List<DebugScoreRecord> scoreRecordsList = new ArrayList<>();
+    return scoreRecordsList;
+  }
+  
+  public ArrayList<MulticlassScoringResult> scoreItemMulticlass(FeatureVector combinedItem, MulticlassScoringOptions options) {
+    ArrayList<MulticlassScoringResult> results = new ArrayList<>();
+    
+    return results;
+  }
+
+  public void save(BufferedWriter writer) throws IOException {
+    ModelHeader header = new ModelHeader();
+    header.setModelType("full_rank_linear");
+    long count = 0;
+    for (Map.Entry<String, Map<String, FloatVector>> familyMap : weightVector.entrySet()) {
+      for (Map.Entry<String, FloatVector> feature : familyMap.getValue().entrySet()) {
+        count++;
+      }
+    }
+    header.setNumRecords(count);
+    header.setLabelDictionary(labelDictionary);
+    ModelRecord headerRec = new ModelRecord();
+    headerRec.setModelHeader(header);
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/airbnb/aerosolve/core/models/FullRankLinearModel.java
```diff
@@ -0,0 +1,164 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+// A full rank linear model that supports multi-class classificaiton.
+// The class vector Y = W' X where X is the feature vector.
+public class FullRankLinearModel extends AbstractModel {
```
deerzq: it seems this is mainly for multi-class classification, why use the name `FullRankLinearModel`

hectorgon: Because the next model type planned is low rank linear model

deerzq: Oh, I see. Then it might make sense to explain what do we mean by 'full rank' in the comment

core/src/main/java/com/airbnb/aerosolve/core/models/FullRankLinearModel.java
```diff
@@ -0,0 +1,164 @@
...
+    weightVector = new HashMap<>();
+    labelDictionary = new ArrayList<>();
+  }
+
+  // In the binary case this is just the score for class 0.
+  @Override
+  public float scoreItem(FeatureVector combinedItem) {
+    Map<String, Map<String, Double>> flatFeatures = Util.flattenFeature(combinedItem);
+    FloatVector sum = scoreFlatFeature(flatFeatures);
+    return sum.values[0];
```
deerzq: for multiclass case this is also the score for class 0, to avoid misuse of this method, perhaps add a warning if there are more than two classes?

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
```diff
@@ -0,0 +1,220 @@
...
+  def train(sc : SparkContext,
+            input : RDD[Example],
+            config : Config,
+            key : String) : FullRankLinearModel = {
+    val options = parseTrainingOptions(config.getConfig(key))
+
+    val pointwise : RDD[Example] =
+      LinearRankerUtils
+        .makePointwiseFloat(input, config, key)
+        .cache()
```
deerzq: for really large training data, `cache()` might result in out of memory problem

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
```diff
@@ -0,0 +1,220 @@
...
+        loss = config.getString("loss"),
+        iterations = config.getInt("iterations"),
+        rankKey = config.getString("rank_key"),
+        lambda = config.getDouble("lambda"),
+        minCount = config.getInt("min_count")
+    )    
+  }
+  
+  def setupModel(options : FullRankLinearTrainerOptions, pointwise : RDD[Example]) : FullRankLinearModel = {
+    val stats = TrainingUtils.getFeatureStatistics(options.minCount, pointwise)
```
deerzq: basically, we only need the count of samples from each class, i.e. count for the feature that represents label, but this method will compute min, max, variance for all features which seems a lot of waste of computation

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
```diff
@@ -0,0 +1,220 @@
...
+                      options : FullRankLinearTrainerOptions,
+                      model : FullRankLinearModel,
+                      pointwise : RDD[Example]) : Array[((String, String), GradientContainer)] = {
+    val modelBC = sc.broadcast(model)
+    
+    pointwise
+    .mapPartitions(partition => {
+      val model = modelBC.value
+      val labelToIdx = model.getLabelToIndex()
+      val dim = model.getLabelDictionary.size()
```
deerzq: make dim a class variable? it seems we use this at multiple places

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
```diff
@@ -0,0 +1,220 @@
...
+      val gradient = scala.collection.mutable.HashMap[(String, String), GradientContainer]()
+      val weightVector = model.getWeightVector()
+
+      partition.foreach(examples => {
+        val flatFeatures = Util.flattenFeature(examples.example.get(0))
+        val labels = flatFeatures.get(options.rankKey)
+        if (labels != null) {
+          val posLabels = labels.keySet().asScala
```
deerzq: so we are supporting multiple labels in this model

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
```diff
@@ -0,0 +1,220 @@
...
+  }
+  
+  def softmaxGradient(sc : SparkContext,
+                      options : FullRankLinearTrainerOptions,
+                      model : FullRankLinearModel,
+                      pointwise : RDD[Example]) : Array[((String, String), GradientContainer)] = {
+    val modelBC = sc.broadcast(model)
+    
+    pointwise
+    .mapPartitions(partition => {
```
deerzq: why don't do subsample here as in spline model trainer?

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
hectorgon: Yeah with other loss functions as well like pairwise hinge loss later on

core/src/main/java/com/airbnb/aerosolve/core/models/FullRankLinearModel.java
hectorgon: I don't want a million log.errors in the log so I will put it in a comment

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
hectorgon: Done

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
hectorgon: Yeah but this is IO bound so I would rather re-use code than write a new one

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
hectorgon: Done

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
hectorgon: Since the trainer is a singleton object I don't feel comfortable using a shared global variable in case the trainer methods are called elsewhere e.g. by an external pipeline

deerzq:
LGTM!

Issue #75, opened 2016-01-11
combineContextAndItems(Example examples)


randombishop:
While trying to use aerosolve for kaggle airbnb competition, I noticed in com.airbnb.aerosolve.core.transforms.Transformer.java
that only string features are being copied from context to examples.
Is this behavior coded on purpose ?
In general, is the differentiation between context and item features only for memory optimization or does it impact the algorithms ranking task ?
Thanks in advance for any advices to help.

hectorgon:
Oops sorry that was not intended. It is mainly for bandwidth optimization when scoring a single context with multiple items (e.g. a single user profile and hundreds of items they might like). For now please use just the repeated example list to store your features.

hectorgon:
Looks like a pretty easy fix. @christhetree - could you please take a stab at this issue

christhetree:
Cool, I'll take a look.

randombishop:
Great, thanks a lot guys

PR #83, opened 2016-01-25
Copy all features from context to example

christhetree:
Work in progress, just checking if this is on the right track.
@hectorgon @yolken

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/transforms/Transformer.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/transforms/Transformer.java
@@ -5,6 +5,8 @@ import com.airbnb.aerosolve.core.FeatureVector;
 import com.typesafe.config.Config;
 
 import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/airbnb/aerosolve/core/transforms/Transformer.java
```diff
@@ -70,23 +70,47 @@ public void combineContextAndItems(Example examples) {
...
   }
 
   public void addContextToItemAndTransform(Map<String, Set<String>> contextStringFeatures,
+                                           Map<String, Map<String, Double>> contextFloatFeatures,
+                                           Map<String, List<Double>> contextDenseFeatures,
                                            FeatureVector item) {
     Map<String, Set<String>> itemStringFeatures = item.getStringFeatures();
+    Map<String, Map<String, Double>> itemFloatFeatures = item.getFloatFeatures();
+    Map<String, List<Double>> itemDenseFeatures = item.getDenseFeatures();
     if (item.getStringFeatures() == null) {
```
hectorgon: Java objects are references, so you probably want to iterate over each map/map and stuff them into the corresponding target item maps otherwise the item transforms might stomp on the same context feature map since java objects are references.

christhetree:
Feel free to take another look when you have a chance and let me know if you want more robust / separate float and dense feature tests.

core/src/test/java/com/airbnb/aerosolve/core/transforms/ModelTransformsTest.java
```diff
@@ -22,14 +24,21 @@ public FeatureVector makeFeatureVector(String familyName,
...
 
     Map<String, Double> map = new HashMap<>();
     map.put("lat", lat);
     map.put("long", lng);
     floatFeatures.put(familyName, map);
 
+    List<Double> list = new ArrayList<>();
+    list.add(lat);
+    list.add(lng);
+    denseFeatures.put(familyName, list);
```
hectorgon: I would use a different family name for the dense feature just to be clearer

core/src/main/java/com/airbnb/aerosolve/core/transforms/Transformer.java
```diff
@@ -70,23 +72,54 @@ public void combineContextAndItems(Example examples) {
...
       item.setStringFeatures(contextStringFeatures);
     } else if (contextStringFeatures != null) {
       itemStringFeatures.putAll(contextStringFeatures);
     }
+    if (item.getFloatFeatures() == null) {
```
hectorgon: This means all the items share the same context hash map .. probably want to move this inside the else if and create a new hash map instead

christhetree:
Added deep copy to all features.

hectorgon:
:+1: please close issue #75 with this

PR #84, opened 2016-01-26
Add string tokenizer transform

christhetree:
This is nowhere close to being finished, just wanted to push it up so that we can more easily discuss the design of the transform, whether an explicit output should be generated, and whether repeated tokens are desired (since the output is currently a set). We can talk about this tomorrow morning.
@hectorgon @yolken


christhetree:
Adding the following resources:
https://github.com/redpony/cdec/blob/master/corpus/tokenize-anything.sh
https://github.com/redpony/cdec/tree/master/corpus/support
The tokenizer will hopefully be able to achieve the same functionality.

core/src/main/java/com/airbnb/aerosolve/core/transforms/StringTokenizerTransform.java
```diff
@@ -0,0 +1,48 @@
...
+    if (stringFeatures == null) {
+      return;
+    }
+
+    Set<String> feature1 = stringFeatures.get(fieldName1);
+    if (feature1 == null) {
+      return;
+    }
+
+    Set<String> output = Util.getOrCreateStringFeature(outputName, stringFeatures);
```
hectorgon: You want a float feature here and output the sum of times the token occured.
e.g. "the cat sat on the mat" should have the : 2 cat : 1 sat : 1 on : 1 mat : 1

core/src/main/java/com/airbnb/aerosolve/core/transforms/StringTokenizerTransform.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Tokenizes strings
+ * "field1" specifies the key of the feature
+ * "field2" specifies the delimiter or regex used to tokenize
+ */
+public class StringTokenizerTransform extends Transform {
+  private String fieldName1;
+  private String delimiterOrRegex;
+  private String outputName;
+
```
hectorgon: I would also add a boolean to delete field1 or not.

core/src/main/java/com/airbnb/aerosolve/core/transforms/StringTokenizerTransform.java
```diff
@@ -0,0 +1,48 @@
...
+  @Override
+  public void configure(Config config, String key) {
+    fieldName1 = config.getString(key + ".field1");
+    delimiterOrRegex = config.getString(key + ".field2");
```
hectorgon: field2 probably should be renamed "regex"

core/src/main/java/com/airbnb/aerosolve/core/transforms/StringTokenizerTransform.java
```diff
@@ -0,0 +1,48 @@
...
+    Set<String> feature1 = stringFeatures.get(fieldName1);
+    if (feature1 == null) {
+      return;
+    }
+
+    Set<String> output = Util.getOrCreateStringFeature(outputName, stringFeatures);
+
+    for (String string : feature1) {
+      String[] tokenizedString = string.split(delimiterOrRegex);
+      Collections.addAll(output, tokenizedString);
```
hectorgon: You probably want to accumulate counts here rather than just keeping the unique token

hectorgon:
Can you please make a different tokenizer for the ones that Avneesh provided? They seem more special cased for English.

christhetree:
I've pushed up the simple default string tokenizer along with its tests.

core/src/main/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransform.java
```diff
@@ -0,0 +1,64 @@
...
+public class DefaultStringTokenizerTransform extends Transform {
+  private String fieldName1;
+  private String regex;
+  private boolean deleteField1;
+  private String outputName;
+
+  @Override
+  public void configure(Config config, String key) {
+    fieldName1 = config.getString(key + ".field1");
+    regex = config.getString(key + ".field2");
```
hectorgon: Please call field2 "regex"
and field3 "delete"

core/src/main/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransform.java
```diff
@@ -0,0 +1,64 @@
...
+    deleteField1 = config.getBoolean(key + ".field3");
+    outputName = config.getString(key + ".output");
+  }
+
+  @Override
+  public void doTransform(FeatureVector featureVector) {
+    Map<String, Set<String>> stringFeatures = featureVector.getStringFeatures();
+    Map<String, Map<String, Double>> floatFeatures = featureVector.getFloatFeatures();
+    if (stringFeatures == null || floatFeatures == null) {
```
hectorgon: Only if string features is null do you return, you can always create the output column

christhetree:
Tests for DeleteStringFeatureColumnTransform will be added in separate PR.

core/src/main/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransform.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import com.typesafe.config.Config;
+
+/**
+ * Tokenizes strings using a regex
+ * "field1" specifies the key of the feature
+ * "regex" specifies the regex used to tokenize
+ * "field3" specifies whether field1 should be deleted or not
```
hectorgon: obsolete comment should be output

core/src/main/java/com/airbnb/aerosolve/core/transforms/DateTransform.java
```diff
@@ -22,15 +22,14 @@
   @Override
   public void configure(Config config, String key) {
     fieldName1 = config.getString(key + ".field1");
-    dateType = config.getString(key + ".field2");
+    dateType = config.getString(key + ".dateType");
```
hectorgon: This is still camelCased

hectorgon:
:+1:

PR #87, opened 2016-01-28
Normalize float transform

hectorgon:
L2 norm a sparse float feature

@deerzq @christhetree @yolken


yolken:
LGTM

core/src/main/java/com/airbnb/aerosolve/core/transforms/NormalizeFloatFeatureTransform.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.List;
+import java.util.Map;
+
+// L2 normalizes a float feature
+public class NormalizeFloatFeatureTransform extends Transform {
+  private String fieldName1;
+
+  @Override
+  public void configure(Config config, String key) {
+    fieldName1 = config.getString(key + ".field1");
```
deerzq: how about adding `output` so that if user can have the option to output to a different feature family and keep both normalized and unnormalized features

PR #88, opened 2016-01-29
Add delete string feature column transform tests

christhetree:


christhetree:
Adding tests for the DeleteStringFeatureColumnTransform
@hectorgon

hectorgon:
:+1:

PR #89, opened 2016-01-29
1. add leaf method to kdtree model

jq:
2. fix unit test of kdtree model

@deerzq @hectorgon


deerzq:
perhaps try to share more code between `leaf` and `queryBox`, but in general LGTM

jq:
@deerzq Thanks for review updated with shared code

deerzq:
LGTM

PR #90, opened 2016-02-01
Add string tokenization helper transforms

christhetree:


christhetree:
Initial skeleton of the advanced string tokenizer for latin character languages.
Modeled after the following: 
https://github.com/redpony/cdec/blob/master/corpus/tokenize-anything.sh
https://github.com/redpony/cdec/tree/master/corpus/support
@asaluja feel free to make changes as you see fit

hectorgon:
Don't you want the replacement maps to be configurable by the config file?

christhetree:
@hectorgon that's another possibility, if we go down that route, I think it would make more sense to just make a "ReplaceStringWithRegexTransform" (where you pass in all the replacements via the config file) and a "NormalizeStringToUtf8Transform" (where you pass in what UTF-8 standard is desired via the config file) and then those two could just be chained together with the DefaultStringTokenizer in whatever order is desired in order to simulate the advanced string tokenizer that @asaluja  is looking for. By generalizing the transforms, additional tokenizers for different languages could also be built more easily in the future by combining these three transforms together in any desired order.

asaluja:
@hectorgon @christhetree some thoughts on this PR.  Overall, breaking the overall tokenization transform into a series of smaller transforms and composing is the way to go.  That way, we can re-use code across different languages, and also create transforms on the fly for new languages (e.g., Chinese word segmentation). 

Desiderata:
* separate lower-case transform: some languages don't need this
* transform to eliminate [stop words](https://en.wikipedia.org/wiki/Stop_words) (would need to rely on an external stop words list, which differs for different languages and domains)
* transform should also be able to take in a list of tokens that *should not* be tokenized (e.g., I.B.M.)
* re: UTF-8 normalization as in [this link](https://docs.oracle.com/javase/tutorial/i18n/text/normalizerapi.html), we should probably stick to NFC, don't really see a need for the others so no need to provide flexibility on this front. 
* can ignore any mention of SGML/XML handling (not our use case), 

Referring to the code in [this tokenizer](https://github.com/redpony/cdec/blob/master/corpus/support/tokenizer.pl), the [Stanford tokenizer](https://github.com/stanfordnlp/CoreNLP/blob/master/src/edu/stanford/nlp/process/PTBTokenizer.java) seems to implement significantly the same functionality, but I am not as familiar with the codebase there, so if that's easier for you to follow we can do that. One thing to note is that the Stanford tokenizer is optimized towards English, while the cdec tokenizer is relatively more language-agnostic. 

I also have a bunch of test cases in mind (particularly, testing the conditions on lines 460-520 in the [cdec tokenizer](https://github.com/redpony/cdec/blob/master/corpus/support/tokenizer.pl) ).  When this transform is done, please tag me on the PR for the test cases and I can help in developing them.  Hope this helps!

christhetree:
@hectorgon is there any way to get a map<String, String> as an input from the config file?
@asaluja added some of the helper transforms for an advanced tokenizer that we may need later; I'm still checking out the Stanford tokenizer, we'll have to discuss reproducing it with transforms vs. potentially integrating its source code in directly

christhetree:
Tests will be added later once the code is finalized
@hectorgon Feel free to take another look when you have a chance and let me know if there's a cleaner way to write the bigram logic / if you had something different in mind
@asaluja These three helper transforms should allow you to implement the Perl tokenizer you were referring to earlier by simply feeding in all the desired regexes in the config file and building a tokenizer pipeline using the list transform. However, the order of the tokens is unfortunately still not preserved due to the Map nature of the string features where order is arbitrary to begin with. We can discuss this more in person.

hectorgon:
@christhetree I try to avoid external IO whevever possible because I don't want to suck in S3 dependencies or what not in serving. If the files are small enough you could embed them in a config like this https://github.com/airbnb/aerosolve/blob/master/core/src/main/java/com/airbnb/aerosolve/core/transforms/KDTreeTransform.java or if they are too large you can have them as a separate java resource that you point to in the config and the user supplies in the jar.

hectorgon:
@deerzq and @askeys for PR comments

core/src/test/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransformTest.java
```diff
@@ -60,7 +61,7 @@ public void testEmptyFeatureVector() {
```
hectorgon: Please add a test to check some of the bigrams

hectorgon:
I have no idea how the tokenizing and regex will interact with UTF8 @asaluja do you know when UTF8 normalization happens and what will happen if the tokenizing happens after UTF8 normalization?

asaluja:
@hectorgon UTF8 normalization should happen first, tokenization should happen afterwards.

christhetree:
@hectorgon Bigram test has been added

hectorgon:
:+1: assuming you are adding the tests for regex and normalize in the PR with the filter stopword transformer

PR #92, opened 2016-02-04
Hinge loss for full rank linear model

hectorgon:
@deerzq @asaluja hinge loss for full rank linear model


demo/twenty_news/README.md
```diff
@@ -71,50 +71,50 @@ Next you can test the performance of the trained model on the training data and
...
-16/01/27 16:57:53 INFO TwentyNewsPipeline$: (TRAIN_PRECISION@2,0.7759964373190826)
-16/01/27 16:57:53 INFO TwentyNewsPipeline$: (TRAIN_PRECISION@20,1.0)
-16/01/27 16:57:53 INFO TwentyNewsPipeline$: (TRAIN_PRECISION@3,0.8075038966822534)
-16/01/27 16:57:53 INFO TwentyNewsPipeline$: (TRAIN_PRECISION@4,0.8268203072812291)
-16/01/27 16:57:53 INFO TwentyNewsPipeline$: (TRAIN_PRECISION@5,0.8409040302827878)
-16/01/27 16:57:53 INFO TwentyNewsPipeline$: (TRAIN_PRECISION@6,0.8517034068136272)
-16/01/27 16:57:53 INFO TwentyNewsPipeline$: (TRAIN_PRECISION@7,0.8604431084391004)
-16/01/27 16:57:53 INFO TwentyNewsPipeline$: (TRAIN_PRECISION@8,0.8675684702738811)
-16/01/27 16:57:53 INFO TwentyNewsPipeline$: (TRAIN_PRECISION@9,0.8743041638833222)
+16/02/04 02:00:19 INFO TwentyNewsPipeline$: (HOLD_ALL_PAIRS_HINGE_LOSS,15.741044936394726)
```
deerzq: great improvement!

training/src/test/scala/com/airbnb/aerosolve/training/FullRankLinearModelTest.scala
```diff
@@ -43,6 +43,11 @@ class FullRankLinearModelTest {
     testFullRankLinear("softmax", 0.8)
   }
 
+  @Test
+  def testFullRankLinearHinge() = {
+    testFullRankLinear("hinge", 0.8)
```
deerzq: did you test higher threshold here? since 'hinge' is doing better, I'd like to see whether it is also improving this simple testing example

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
```diff
@@ -175,6 +176,73 @@ object FullRankLinearTrainer {
...
+          val labels = flatFeatures.get(options.rankKey)
+          if (labels != null && labels.size() > 0) {
+            val posLabels = labels.toArray
+            // Pick a random positive label
+            val posLabelRnd = rnd.nextInt(posLabels.size)
+            val (posLabel, posMargin) = posLabels(posLabelRnd)
+            val posIdx = labelToIdx.get(posLabel)
+            // Pick a random other label
+            var negIdx = rnd.nextInt(dim)
+            while (negIdx == posIdx) {
```
deerzq: If there are multiple possible posLabel, `negIdx!=posIdx` does not imply it is not posLabel, perhaps need to check all possible pos label idx

training/src/test/scala/com/airbnb/aerosolve/training/FullRankLinearModelTest.scala
deerzq: Shall we also test multi-class, multi-label case?

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
```diff
@@ -175,6 +176,73 @@ object FullRankLinearTrainer {
...
+    pointwise
+      .mapPartitions(partition => {
+        val model = modelBC.value
+        val labelToIdx = model.getLabelToIndex()
+        val dim = model.getLabelDictionary.size()
+        val gradient = scala.collection.mutable.HashMap[(String, String), GradientContainer]()
+        val weightVector = model.getWeightVector()
+        val rnd = new Random()
+
+        partition.foreach(examples => {
```
deerzq: partition consists of multiply examples, but in partition.foreach we are only looking at one example, so perhaps rename this as `example`

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
```diff
@@ -175,6 +176,73 @@ object FullRankLinearTrainer {
...
+            val posLabelRnd = rnd.nextInt(posLabels.size)
+            val (posLabel, posMargin) = posLabels(posLabelRnd)
+            val posIdx = labelToIdx.get(posLabel)
+            // Pick a random other label
+            var negIdx = rnd.nextInt(dim)
+            while (negIdx == posIdx) {
+              negIdx = rnd.nextInt(dim)
+            }
+            val negLabel = model.getLabelDictionary.get(negIdx).label
+            val negMargin : Double = if (labels.containsKey(negLabel)) labels.get(negLabel) else 0.0
```
deerzq: I am very confused here. I thought for one example, labels would contain positive labels for this example

hectorgon: A negative could be a positive label with a smaller margin or not a positive label in which case the margin is 0.0

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
hectorgon: It's ok I just need another label not necessarily pos vs only negative but pos with higher margin vs pos with smaller margin

training/src/test/scala/com/airbnb/aerosolve/training/FullRankLinearModelTest.scala
hectorgon: They both get 100% correct just lowering it to make sure the test is stable. I will add a multi-label case later

training/src/main/scala/com/airbnb/aerosolve/training/FullRankLinearTrainer.scala
deerzq: oh, I see, perhaps add some comments in the code?

hectorgon: Done

deerzq:
LGTM

PR #97, opened 2016-02-10
Make the L2 gradient update be part of the rprop

hectorgon:
The L2 regularization term should have been part of the rprop procedure instead of happening after the steps
@deerzq @Patrick1860


Patrick1860:
LGTM I was a bit unclear why the regularization term is out of grad before. Should have caught it out earlier

deerzq:
LGTM

askeys:
:+1:

PR #98, opened 2016-02-15
add debugScoreComponents in full rank model

deerzq:
to: @hectorgon @jq 

@jq You can add the `debugScoreItem` in FullRankModel as a practice: https://github.com/airbnb/aerosolve/blob/master/core/src/main/java/com/airbnb/aerosolve/core/models/FullRankLinearModel.java#L58


jq:
LGTM

PR #99, opened 2016-02-16
Low rank linear model

deerzq:
Define low rank linear model, implement the scoring, saving and loading part.

TODO: implement the model trainer.

to: @hectorgon @jq

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/LowRankLinearModel.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/LowRankLinearModel.java
@@ -18,25 +18,45 @@ import com.airbnb.aerosolve.core.util.FloatVector;
 import lombok.Getter;
 import lombok.Setter;
 
-// A full rank linear model that supports multi-class classification.
-// The class vector Y = W' X where X is the feature vector.
-// It is full rank because the matrix W is num-features by num-labels in dimension.
-// Use a low rank model if you want better generalization.
-public class FullRankLinearModel extends AbstractModel {
+// A low rank linear model that supports multi-class classification.
+// The class vector y = W' * V * x where x is d-dim feature vector.
+// Suppose we have Y different labels and the D is the dimension of the joint feature-label space
+// V: D-by-d matrix, mapping from feature space to the joint embedding
+// W: D-by-Y matrix, mapping from label space to the joint embedding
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/airbnb/aerosolve/core/models/LowRankLinearModel.java
```diff
@@ -0,0 +1,223 @@
...
+  private ArrayList<LabelDictionaryEntry> labelDictionary;
+
+  @Getter
+  @Setter
+  private Map<String, Integer> labelToIndex;
+
+  // size of the embedding
+  @Getter
+  @Setter
+  private int D;
```
hectorgon: I would use something other than D, maybe embeddingDimension?

core/src/main/java/com/airbnb/aerosolve/core/models/LowRankLinearModel.java
```diff
@@ -0,0 +1,223 @@
...
+
+    return results;
+  }
+
+  public FloatVector scoreFlatFeature(Map<String, Map<String, Double>> flatFeatures) {
+    int dim = labelDictionary.size();
+    FloatVector sum = new FloatVector(dim);
+    FloatVector fvProjection = new FloatVector(D);
+
+    // compute the projection from feature space to D-dim joint space
```
hectorgon: I would break these components out into different function calls because they are useful in the future.
e.g. projectFeature.

hectorgon:
Minor comments but the skeleton looks good :+1:

deerzq:
@hectorgon Thanks for the review, just updated the PR.

core/src/test/java/com/airbnb/aerosolve/core/models/LowRankLinearModelTest.java
```diff
@@ -0,0 +1,265 @@
...
+      }
+      case "color": {
+        feature.put("red", 2.0);
+        feature.put("black", 4.0);
+        floatFeatures.put("c", feature);
+        break;
+      }
+      case "fruit": {
+        feature.put("apple", 1.0);
+        feature.put("kiwi", 3.0);
```
jq: how about test feature.put("apple", 1.0); and feature.put("black", 4.0); together?

jq:
LGTM

PR #100, opened 2016-02-17
low rank model trainer

deerzq:
Trainer for the low-rank model

@hectorgon

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
@@ -1,0 +1,329 @@
+package com.airbnb.aerosolve.training
+
+import com.airbnb.aerosolve.core.{Example, LabelDictionaryEntry}
+import com.airbnb.aerosolve.core.models.LowRankLinearModel
+import com.airbnb.aerosolve.core.util.FloatVector
+import com.airbnb.aerosolve.core.util.Util
+import com.airbnb.aerosolve.training.GradientUtils._
+import com.typesafe.config.Config
+import org.slf4j.{LoggerFactory, Logger}
+import org.apache.spark.SparkContext
+import org.apache.spark.SparkContext._
+import org.apache.spark.rdd.RDD
+
+import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
+import scala.collection.immutable.HashMap
+import scala.util.{Try, Random}
+
+/*
+ * A trainer that generates a low rank linear model.
+ *
+ * */
+
+object LowRankLinearTrainer {
+  private final val log: Logger = LoggerFactory.getLogger("LowRankLinearTrainer")
+  private final val LABEL_EMBEDDING_KEY = "$label_embedding"
+  case class LowRankLinearTrainerOptions(loss : String,
+                                         iterations : Int,
+                                         rankKey : String,
+                                         lambda : Double,
+                                         subsample : Double,
+                                         minCount : Int,
+                                         cache : String,
+                                         solver : String,
+                                         embeddingDimension : Int,
+                                         rankLossType: String)
+
+  def train(sc : SparkContext,
+            input : RDD[Example],
+            config : Config,
+            key : String) : LowRankLinearModel = {
+    val options = parseTrainingOptions(config.getConfig(key))
+
+    val raw : RDD[Example] =
+      LinearRankerUtils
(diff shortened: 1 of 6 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
```diff
@@ -0,0 +1,329 @@
...
+    val pointwise = options.cache match {
+      case "memory" => raw.cache()
+      case _ : String => raw
+    }
+
+    val model = setupModel(options, pointwise)
+
+    modelIteration(sc, options, model, pointwise)
+
+    filterZeros(model)
```
hectorgon: I don't think you will get zero rows without L1 regularization

training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
```diff
@@ -0,0 +1,329 @@
...
+  private def uniformRankLoss(k: Int, dim: Int): Float = {
+    // dim is the number of classes
+    // assume we always have dim > 1
+    k * 1.0f / (dim - 1.0f)
+  }
+
+  private def nonUniformRankLoss(k: Int, dim: Int): Float = {
+    var loss = 0.0f
+    for (i <- 1 to k) {
+      loss += 1.0f / i
```
hectorgon: For efficiency this can probably be computed once and passed as an array

training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
```diff
@@ -0,0 +1,329 @@
...
+        entry.setCount(kv._2.count.toInt)
+        dict.add(entry)
+      } else {
+        if (!featureWeights.containsKey(family)) {
+          featureWeights.put(family, new java.util.HashMap[java.lang.String, FloatVector]())
+        }
+        val familyMap = featureWeights.get(family)
+        if (!familyMap.containsKey(feature)) {
+          count = count + 1
+          familyMap.put(feature, FloatVector.getGaussianVector(embeddingSize))
```
hectorgon: Ideally this should be uniformly drawn from +/- 1/sqrt(embeddingSize)
http://stats.stackexchange.com/questions/47590/what-are-good-initial-weights-in-a-neural-network

hectorgon:
LGTM with some minor comments regarding initialization and L1 regularization

deerzq:
@hectorgon since we are not going to use sparse boost solver, perhaps we don't need to compute the real norm in GradientContainer? this can save some computation

hectorgon:
Yes no need to use the real norm that was just for debugging anyway. One good check for convergence is to pick a random feature (or label) and print the top K labels close to that feature/label.
e.g. for twenty news alt.atheism should be near alt.religion. This is optional mostly for debugging. One could also supply a list of labels to compute nearest neighbors to for debugging for convergence at each iteration

training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
deerzq: Thanks, updated the PR to use uniform sampling for initialization

deerzq:
removed the feature norm computation part since it's not used in rprop solver

training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
deerzq: added a `todo` for this task in the code

training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
```diff
@@ -0,0 +1,327 @@
...
+    }
+  }
+
+  private def uniformRankLoss(k: Int, dim: Int): Float = {
+    // dim is the number of classes
+    // assume we always have dim > 1
+    k * 1.0f / (dim - 1.0f)
+  }
+
+  private def nonUniformRankLoss(k: Int, dim: Int): Float = {
```
hectorgon: I think this is called "weighted approximate ranking loss" or WARP loss.

deerzq: this term is defined in the wsabie paper, I guess all three losses are 'WARP' loss.

PR #101, opened 2016-02-19
Misc updates for low rank model trainer

deerzq:
- added unit test for multi-label cases
- fix gradient updates
- normalizing weights
- add a simple gradient descent method
to: @hectorgon @Patrick1860

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/GradientUtils.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/GradientUtils.scala
@@ -62,12 +62,12 @@ object GradientUtils {
             step : scala.collection.mutable.HashMap[(String, String), FloatVector],
             weightVector : java.util.Map[String,java.util.Map[String,com.airbnb.aerosolve.core.util.FloatVector]],
             dim :Int,
-            lambda : Double) = {
+            lambda : Double,
+            deltaMax : Float = 50.0f) = {
     // Some fixed parameters for rprop, they don't matter much as the algorithm adapts them
     // so just keep them fixed here.
     // Step size parameters
     val deltaInitial = 0.01f
-    val deltaMax = 50.0f
     val deltaMin = 1e-6f
     // Change in step size parameters
     val etaPlus = 1.2f
--- a/training/src/main/scala/com/airbnb/aerosolve/training/GradientUtils.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/GradientUtils.scala
@@ -63,7 +63,7 @@ object GradientUtils {
             weightVector : java.util.Map[String,java.util.Map[String,com.airbnb.aerosolve.core.util.FloatVector]],
             dim :Int,
             lambda : Double,
-            deltaMax : Float = 50.0f) = {
+            deltaMax : Float = 1.0f) = {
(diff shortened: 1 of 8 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
```diff
@@ -139,7 +153,7 @@ object LowRankLinearTrainer {
             do {
               // Pick a random other label
               val (idx, label, margin, iter) = pickRandomOtherLabel(model, labels, posIdx, posMargin, rnd, dim)
-              if (iter < dim) {
+              if (iter < dim * 2) {
```
hectorgon: I think this messes up the warp loss

training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
```diff
@@ -217,6 +230,30 @@ object LowRankLinearTrainer {
     TrainingUtils.saveModel(model, config, key + ".model_output")
   }
 
+  private def normalizeWeightVectors(model: LowRankLinearModel, maxNorm: Double) = {
```
hectorgon: capWeightVector? but this should ideally call a capNorm function in FloatVector so you don't duplicate code twice

training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
deerzq: fixed

training/src/main/scala/com/airbnb/aerosolve/training/LowRankLinearTrainer.scala
deerzq: done

deerzq:
@hectorgon removed the multilabel test since it's very flaky; make the delta_max in rprop an input parameter, since for Low rank model, we do the weight normalization, it does not make sense to allow too large step.

training/src/main/scala/com/airbnb/aerosolve/training/GradientUtils.scala
```diff
@@ -109,7 +109,30 @@ object GradientUtils {
     })
     val stepNorms = step.map(kv => math.sqrt(kv._2.dot(kv._2)))
     log.info("Sum of Step L2 norms = " + stepNorms.sum)
+    log.info("Average of Step L2 norms = " + stepNorms.sum / stepNorms.size)
+    log.info("Max of Step L2 norms = " + stepNorms.max)
     val median = stepNorms.toBuffer.sorted.get(stepNorms.size / 2)
     log.info("Median Step L2 norms = " + median)
   }
+
+  def gradientDescent(gradients : Map[(String, String), GradientContainer],
```
hectorgon: This doesn't seem to be used are you keeping it just in case?

hectorgon:
:+1:

training/src/main/scala/com/airbnb/aerosolve/training/GradientUtils.scala
deerzq: yes, just leave it there. It might be useful in the future.

PR #102, opened 2016-02-20
Multiclass decision tree

hectorgon:
@deerzq @yolken @asaluja @Patrick1860 @spencerde

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
@@ -344,9 +344,11 @@ object DecisionTreeTrainer {
       val response = BoostedStumpsModel.getStumpResponse(candidateOpt.get, example)
       for (kv <- example.get(rankKey).asScala) {
         if (response) {
-          rightDist.put(kv._1, kv._2)
+          val v = rightDist.getOrElse(kv._1, 0.0)
+          rightDist.put(kv._1, kv._2 + v)
         } else {
-          leftDist.put(kv._1, kv._2)
+          val v = rightDist.getOrElse(kv._1, 0.0)
+          leftDist.put(kv._1, kv._2 + v)
         }
       }
     }
--- a/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
@@ -19,7 +19,7 @@ import scala.collection.JavaConverters._
 
 // Types of split criteria
 object SplitCriteriaTypes extends Enumeration {
-  val Classification, Regression = Value
+  val Classification, Regression, Multiclass = Value
 }
 
 // The decision tree is meant to be a prior for the spline model / linear model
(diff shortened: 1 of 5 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
```diff
@@ -289,6 +317,64 @@ object DecisionTreeTrainer {
...
+    val scale = 1.0 / (sum * sum)
+    var impurity : Double = 0.0
+    for (kv1 <- dist) {
+      for (kv2 <- dist) {
+        if (kv1._1 != kv2._1) {
+          impurity += scale * kv1._2 * kv2._2
+        }
+      }
+    }
+    impurity
```
deerzq: multiply `scale` at the end instead of at line327 to save some computation?

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
```diff
@@ -289,6 +317,64 @@ object DecisionTreeTrainer {
...
+                              splitCriteria : String,
+                              candidateOpt : Option[ModelRecord]): Option[Double] = {
+    val leftDist = scala.collection.mutable.HashMap[String, Double]()
+    val rightDist = scala.collection.mutable.HashMap[String, Double]()
+
+    for (example <- examples) {
+      val response = BoostedStumpsModel.getStumpResponse(candidateOpt.get, example)
+      for (kv <- example.get(rankKey).asScala) {
+        if (response) {
+          rightDist.put(kv._1, kv._2)
```
deerzq: I feel the value should be accumulated, otherwise this will be determined by the last example only. how about something like?
`val v = rightDist.getOrElse(kv._1)`

`rightDist.put(kv._1, v + kv._2)`

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
```diff
@@ -289,6 +317,64 @@ object DecisionTreeTrainer {
...
+        } else {
+          leftDist.put(kv._1, kv._2)
```
deerzq: same here

training/src/test/scala/com/airbnb/aerosolve/training/DecisionTreeModelTest.scala
```diff
@@ -44,6 +44,16 @@ class DecisionTreeModelTest {
   }
 
   @Test
+  def testDecisionTreeTrainerMulticlassHellinger() = {
+    testDecisionTreeMulticlassTrainer("multiclass_hellinger", 0.5)
```
deerzq: higher threshold?

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
hectorgon: Maps are unique associative containers? I don't think it is possible to have more than one value per key in the label map?

training/src/test/scala/com/airbnb/aerosolve/training/DecisionTreeModelTest.scala
hectorgon: The randomness makes the accuracy jump up and down and could make the test flaky. Since random is 0.25 I figured twice of random should be a good cutoff

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
deerzq: `v + kv._2` is just one value, not two values.

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
hectorgon: Done

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
hectorgon: Ah OK good catch fixed

deerzq:
LGTM!

PR #103, opened 2016-02-22
Add “In the Wild” section

ljharb:
The idea is that companies and/or projects can list themselves here, just like https://github.com/airbnb/javascript#in-the-wild.

It's in a separate file so that PRs to it don't have to touch the main README.


hectorgon:
:+1:

PR #104, opened 2016-02-22
Add string tokenizer helper transforms test

christhetree:


christhetree:
@hectorgon @asaluja feel free to take a look when you have a chance
The filter stopword transform is not necessary since it should be easily implementable using the ReplaceAllStringsTransform. Also, let me know if you still want a split on space test for the normalize utf8 transform, I played around with that a bit, but couldn't get the desired behavior (it always split correctly).

hectorgon:
:+1:

PR #106, opened 2016-02-24
MLP Model

deerzq:
Add a simple MLP model with one output node

todo: add the trainer for MLP model

to: @hectorgon @christhetree 
cc: @Patrick1860 @jq

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/MlpModel.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/MlpModel.java
@@ -1,0 +1,326 @@
+package com.airbnb.aerosolve.core.models;
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.io.Serializable;
+import java.util.*;
+
+import com.airbnb.aerosolve.core.DebugScoreRecord;
+import com.airbnb.aerosolve.core.FeatureVector;
+import com.airbnb.aerosolve.core.ModelHeader;
+import com.airbnb.aerosolve.core.ModelRecord;
+import com.airbnb.aerosolve.core.util.Util;
+import com.airbnb.aerosolve.core.util.FloatVector;
+import lombok.Getter;
+import lombok.Setter;
+
+
+/**
+ * Multilayer perceptron (MLP) model https://en.wikipedia.org/wiki/Multilayer_perceptron
+ * The current implementation is for the case where there is only one output node.
+ */
+public class MlpModel extends AbstractModel {
+
+  private static final long serialVersionUID = -6870862764598907090L;
+
+  // weights that define the projection from input layer (layer0) to the first hidden layer
+  // or the output layer
+  @Getter
+  @Setter
+  private Map<String, Map<String, FloatVector>> inputLayerWeights;
+
+  // if there is hidden layer, this defines the projection
+  // from one hidden layer to the next hidden layer or output layer
+  // map from hidden layer id to hidden layer weights, id starts from 0
+  @Getter
+  @Setter
+  private Map<Integer, ArrayList<FloatVector>> hiddenLayerWeights;
+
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/airbnb/aerosolve/core/models/MlpModel.java
```diff
@@ -0,0 +1,326 @@
...
+
+  // number of layers (excluding input layer and output layer)
+  @Getter
+  @Setter
+  private int numHiddenLayers;
+
+  // number of nodes for each hidden layer and output layer (does not include input layer)
+  @Getter
+  @Setter
+  private Map<Integer, Integer> layerNodeNumber;
```
hectorgon: Isn't just an array sufficient

core/src/main/java/com/airbnb/aerosolve/core/models/MlpModel.java
```diff
@@ -0,0 +1,326 @@
...
+    layerActivations = new HashMap<>();
+
+    for (int i = 0; i < numHiddenLayers; i++) {
+      int nodeNum = nodeNumbers.get(i);
+      layerNodeNumber.put(i, nodeNum);
+      hiddenLayerWeights.put(i, new ArrayList<>(nodeNum));
+      bias.put(i, new FloatVector(nodeNum));
+      layerActivations.put(i, new FloatVector(nodeNum));
+    }
+    // output layer has one node
```
hectorgon: This is not the case for multiclass right?

core/src/main/java/com/airbnb/aerosolve/core/util/FloatVector.java
```diff
@@ -103,6 +103,22 @@ public void capNorm(float maxNorm) {
     }
   }
 
+  public void sigmoid() {
+    for (int i = 0; i < values.length; i++) {
+      values[i] = 1.0f /(1.0f + (float) Math.exp(-values[i]));
```
hectorgon: Might want to check if values is smaller than -20 or so in case of overflow.

core/src/main/java/com/airbnb/aerosolve/core/util/FloatVector.java
```diff
@@ -103,6 +103,22 @@ public void capNorm(float maxNorm) {
...
+    }
+  }
+
+  public void tanh() {
+    for (int i = 0; i < values.length; i++) {
+      values[i] = 2.0f /(1.0f + (float) Math.exp(-2.0 * values[i])) - 1.0f;
```
hectorgon: Java has a Math.tanh

core/src/main/thrift/MLSchema.thrift
```diff
@@ -76,6 +76,12 @@ struct ModelHeader {
   // Multiclass labels.
   7: optional list<LabelDictionaryEntry> labelDictionary;
   8: optional map<string, list<double>> labelEmbedding;
+  // The number of hidden layers in neural network
+  9: optional i32 numHiddenLayers;
+  // the activation function in neural network
+  10: optional string activationFunction;
```
hectorgon: Shouldn't this be per layer? I think Model record has functionalForm already

core/src/main/java/com/airbnb/aerosolve/core/models/MlpModel.java
deerzq: yes, this is just for binary classification and regression. I may add the multi-class case later

core/src/main/thrift/MLSchema.thrift
deerzq: I am thinking to start with a global activation function, but I can change this to support per-layer activation function

deerzq:
@hectorgon thanks for review, updated the PR to address your comments, PTAL

christhetree:
@deerzq The only thing that jumped out at me was the sigmoid overflowing which Hector also pointed out, so LGTM so far

hectorgon:
:+1: but please make sure it is easy to upgrade to multiclass in the future

PR #109, opened 2016-02-26
Add MLP model trainer

deerzq:
Initial version of the MLP trainer with simple unit tests. will add more tests in next PR

to: @hectorgon @christhetree 
cc: @jq

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
@@ -12,10 +12,11 @@ import org.apache.spark.SparkContext._
 import org.apache.spark.rdd.RDD
 
 import scala.collection.JavaConversions._
-import scala.util.{Try, Random}
+import scala.util.Try
 
 /**
   * A trainer that generates a MLP model.
+  * TODO (Peng): add maxNorm and weightDecay regularizations
   */
 
 object MlpModelTrainer {
--- a/training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
@@ -16,7 +16,7 @@ import scala.util.Try
 
 /**
   * A trainer that generates a MLP model.
-  * TODO (Peng): add maxNorm and weightDecay regularizations
+  * TODO (Peng): add maxNorm regularizations
   */
 
 object MlpModelTrainer {
@@ -130,6 +130,7 @@ object MlpModelTrainer {
           }
           val grad = options.loss match {
             case "hinge" => computeHingeGradient(score, fv, options)
+            case "regression" => computeRegressionGradient(score, fv, options)
             case _ => computeHingeGradient(score, fv, options)
           }
           // back-propagation for updating gradient
@@ -140,13 +141,18 @@ object MlpModelTrainer {
           // activation: the output of a node
           val outputNodeDelta = computeActivationGradient(score, func) * grad
 
-          backPropagation(model, outputNodeDelta.toFloat, gradient, flatFeatures)
+          backPropagation(model, outputNodeDelta.toFloat, gradient, flatFeatures, options.weightDecay.toFloat)
         })
         gradient.iterator
       })
+      .mapValues(fv => (fv, 1.0))
       .reduceByKey((a, b) => {
-        a.add(b)
-        a
+        a._1.add(b._1)
+        (a._1, a._2 + b._2)
+      })
+      .mapValues(x => {
+        x._1.scale(1.0f / x._2.toFloat)
+        x._1
       })
       .collectAsMap
       .toMap
@@ -155,7 +161,8 @@ object MlpModelTrainer {
   def backPropagation(model: MlpModel,
                       outputNodeDelta: Float,
                       gradient: scala.collection.mutable.HashMap[(String, String), FloatVector],
-                      flatFeatures: java.util.Map[String, java.util.Map[java.lang.String, java.lang.Double]]) = {
+                      flatFeatures: java.util.Map[String, java.util.Map[java.lang.String, java.lang.Double]],
+                      weightDecay: Float = 0.0f) = {
     // outputNodeDelta: gradient of the loss function w.r.t the input of the output node
     val numHiddenLayers = model.getNumHiddenLayers
     val layerNodeNumber = model.getLayerNodeNumber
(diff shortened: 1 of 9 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
```diff
@@ -0,0 +1,429 @@
...
+      case "memory" => raw.cache()
+      case _ : String => raw
+    }
+
+    val model = setupModel(trainerOptions, networkOptions, pointwise)
+
+    modelIteration(sc, trainerOptions, model, pointwise)
+
+    trainerOptions.cache match {
+      case "memory" => pointwise.unpersist()
```
hectorgon: case _ => Unit or else it will crash

training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
```diff
@@ -0,0 +1,429 @@
...
+  private def computeActivationGradient(activation: Double,
+                                        func: FunctionForm): Double = {
+    // compute the gradient of activation w.r.t input
+    // note: input only need to be specified for RELU
+    func match {
+      case FunctionForm.SIGMOID => activation * (1.0 - activation)
+      case FunctionForm.RELU => if (activation > 0) 1.0 else 0.0
+      case FunctionForm.IDENTITY => 1.0
+      case FunctionForm.TANH => 1.0 - activation * activation
+      case _ => activation * (1.0 - activation)
```
hectorgon: I would assert fail since this is a training function

deerzq:
Just added some unit tests.

training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
```diff
@@ -0,0 +1,467 @@
...
+  }
+
+  private def computeUpdates(prevUpdate: FloatVector,
+                             momentum: Float,
+                             learningRate: Float,
+                             gradient: FloatVector): FloatVector = {
+    // based on hinton's dropout paper: http://arxiv.org/pdf/1207.0580.pdf
+    val update: FloatVector = new FloatVector(prevUpdate.length)
+    update.multiplyAdd(momentum, prevUpdate)
+    update.multiplyAdd(-(1.0f - momentum) * learningRate, gradient)
```
Patrick1860: reference to this formula?  I saw one version is mu * v - learning_rate * dx which may transfer to update.multiplyAdd(momentum, prevUpdate)
update.multiplyAdd(-(1.0f ) * learningRate, gradient)

Patrick1860: reference to this formula?  I saw one version is mu * v - learning_rate * dx which transfers to update.multiplyAdd(momentum, prevUpdate)
update.multiplyAdd(-(1.0f ) * learningRate, gradient)

training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
```diff
@@ -0,0 +1,467 @@
...
+    for (i <- (0 until numHiddenLayers).reverse) {
+      // i decreases from numHiddenLayers-1 to 0
+      val numNode = layerNodeNumber.get(i)
+      val numNodeUpperLayer = layerNodeNumber.get(i + 1)
+      val func = activationFunctions.get(i)
+      val thisLayerDelta = new FloatVector(numNode)
+      // compute gradient of weights from the i-th layer to the (i+1)-th layer
+      val activations = model.getLayerActivations.get(i)
+      val hiddenLayerWeights = model.getHiddenLayerWeights.get(i)
+      val biasKey = (LAYER_PREFIX + i.toString, BIAS_PREFIX)
```
Patrick1860: not sure if it is more clear to call it UpperLayerBiasKey as it is the input of this layer

training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
```diff
@@ -0,0 +1,467 @@
...
+import org.apache.spark.SparkContext
+import org.apache.spark.SparkContext._
+import org.apache.spark.rdd.RDD
+
+import scala.collection.JavaConversions._
+import scala.util.Try
+
+/**
+  * A trainer that generates a MLP model.
+  * TODO (Peng): add maxNorm regularizations
```
Patrick1860: great~

training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
```diff
@@ -0,0 +1,467 @@
...
+      val numNodeUpperLayer = layerNodeNumber.get(i + 1)
+      val func = activationFunctions.get(i)
+      val thisLayerDelta = new FloatVector(numNode)
+      // compute gradient of weights from the i-th layer to the (i+1)-th layer
+      val activations = model.getLayerActivations.get(i)
+      val hiddenLayerWeights = model.getHiddenLayerWeights.get(i)
+      val biasKey = (LAYER_PREFIX + i.toString, BIAS_PREFIX)
+      val biasGrad = gradient.getOrElse(biasKey, new FloatVector(numNode))
+      for (j <- 0 until numNode) {
+        val key = (LAYER_PREFIX + i.toString, NODE_PREFIX + j.toString)
```
Patrick1860: add some comments here about how the drop out plays here in the backPropagation process? e.g how the gradient of parameters that related to the omitted note set to 0 in this mini batch?

deerzq: the drop out part is implemented in the model, because the activations corresponding to dropped nodes are zeros, the gradient for the weights that connecting that activation with the next node would be 0, so we don' t need to take special care here. will add some comments

deerzq:
@hectorgon updated the PR to address your comments, PTAL? Thanks!

hectorgon:
:+1:  but did you address Pai's comment about the momentum?

deerzq:
@hectorgon there are different implementations for momentum, I am using the formula in http://arxiv.org/pdf/1207.0580.pdf

training/src/main/scala/com/airbnb/aerosolve/training/MlpModelTrainer.scala
Patrick1860: :+1:

Patrick1860:
LGTM didn't see the implementation details of the paper that is hiding after the reference section :).

deerzq:
Thanks for reviewing, will add multi-class part for the trainer and properly test with 20 news.

PR #112, opened 2016-03-09
Allow to specify label min count in FullRankTrainer

deerzq:
There are cases when a particular type of label occurs really rarely, less than the feature min count, the full rank trainer will throw null pointer error. This PR allows user to specify a different min_count value for label 

to: @jq @Patrick1860 
cc: @hectorgon

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
@@ -258,6 +258,40 @@ object TrainingUtils {
       .collect
   }
 
+  def getLabelCounts(minCount : Int,
+                     input : RDD[Example],
+                     rankKey: String) : Array[((String, String), Int)] = {
+    input
+      .mapPartitions(partition => {
+        // family, feature name => count
+        val weights = new ConcurrentHashMap[(String, String), Int]().asScala
+        partition.foreach(examples => {
+          for (i <- 0 until examples.example.size()) {
+            val example = examples.example.get(i)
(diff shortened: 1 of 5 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
```diff
@@ -258,6 +258,40 @@ object TrainingUtils {
       .collect
   }
 
+  def getLabelCounts(minCount : Int,
+                     input : RDD[Example],
+                     rankKey: String) : Array[((String, String), Int)] = {
+    input
+      .mapPartitions(partition => {
```
jq: seems mapPartitions is faster than map, we may want to change most of map into mapPartitions?

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
```diff
@@ -258,6 +258,40 @@ object TrainingUtils {
       .collect
   }
 
+  def getLabelCounts(minCount : Int,
+                     input : RDD[Example],
+                     rankKey: String) : Array[((String, String), Int)] = {
```
jq: rankKey is feature family? how about name it as featureFamily?

deerzq: `rankKey` is the featureFamily name for label, we are using this word many places in Aerosolve. I feel we can keep it this way in this PR. Perhaps do a refactor in the future if necessary

jq:
LGTM

PR #113, opened 2016-03-10
allow eval binary classification on list

deerzq:
When we try to run the evaluation on a lot of break-down dimensions, say thousands of dimensions, this allows us to run the eval in parallel at dimension level.

to: @jq @Patrick1860 @hectorgon

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
@@ -19,7 +19,7 @@ object Evaluation {
   def evaluateBinaryClassification(records : RDD[EvaluationRecord],
                                    buckets : Int,
                                    evalMetric : String) : Array[(String, Double)] = {
-    var metrics  = Buffer[(String, Double)]()
+    var metrics  = mutable.Buffer[(String, Double)]()
     var bestF1 = -1.0
     val thresholds = records.map(x => x.score).histogram(buckets)._1
     // Search all thresholds for the best F1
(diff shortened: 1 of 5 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
```diff
@@ -57,6 +57,48 @@ object Evaluation {
       .toArray
   }
 
+  def evaluateBinaryClassification(records : List[EvaluationRecord],
+                                   buckets : Int,
+                                   evalMetric : String) : Array[(String, Double)] = {
+    var metrics  = mutable.Buffer[(String, Double)]()
+    var bestF1 = -1.0
+    val scores: List[Double] = records.map(x => x.score)
+    val thresholds = getThresholds(records.map(x => x.score), buckets)
```
jq: getThresholds(scores, buckets) ?

training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
```diff
@@ -57,6 +57,48 @@ object Evaluation {
       .toArray
   }
 
+  def evaluateBinaryClassification(records : List[EvaluationRecord],
+                                   buckets : Int,
+                                   evalMetric : String) : Array[(String, Double)] = {
+    var metrics  = mutable.Buffer[(String, Double)]()
+    var bestF1 = -1.0
+    val scores: List[Double] = records.map(x => x.score)
```
jq: how about evaluateBinaryClassification(List[Double]
then we can reuse code?

jq:
LGTM,  share more code if possible

PR #114, opened 2016-03-15
add distance for kdtree node

jq:
@deerzq @hectorgon


core/src/main/java/com/airbnb/aerosolve/core/models/KDTreeModel.java
```diff
@@ -44,6 +45,12 @@ public int leaf(double x, double y) {
     }
   }
 
+  // assume kdtree node x is lat, and y is lng, return distance in miles between kdtree's max and min
+  public double miles(int id) {
```
deerzq: kdTree may not be only used in the lat, lng case, the distance unit may not always be mile. You may want to rename it to something more general like `distance` instead of `mile`

jq: thanks for review!  updated as distanceInMiles

deerzq:
LGTM

PR #119, opened 2016-04-06
Add FloatCrossFloatTransform

christhetree:
to: @jq 
cc: @deerzq 

Content:
This is just a quick foundation of what I envisioned a float cross float transform might look like (essentially a MoveFloatToString transform followed by a StringCrossFloat transform, but without deleting the original float feature or creating an intermediate string feature which would occur if these two transforms were applied back to back). The ability to specify which keys to cross could be added as well instead of crossing the entire family.

Please let me know what you think / if this is what you need. Once we're on the same page I'll add comments and tests. Thanks!

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatCrossFloatTransform.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatCrossFloatTransform.java
@@ -7,6 +7,11 @@ import java.util.Map;
 
 import com.typesafe.config.Config;
 
+/**
+ * Takes the floats in fieldName1, quantizes them into buckets, converts them to strings, then
+ * crosses them with the floats in fieldName2 and then stores the result in a new float feature
+ * output specified by outputName.
+ */
 public class FloatCrossFloatTransform extends Transform {
   private String fieldName1;
   private double bucket;
--- a/core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatCrossFloatTransform.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatCrossFloatTransform.java
@@ -1,0 +1,72 @@
+package com.airbnb.aerosolve.core.transforms;
+
+import com.airbnb.aerosolve.core.FeatureVector;
+import com.airbnb.aerosolve.core.util.Util;
+
+import java.util.Map;
+
+import com.typesafe.config.Config;
+
+public class FloatCrossFloatTransform extends Transform {
+  private String fieldName1;
+  private double bucket;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatCrossFloatTransform.java
```diff
@@ -0,0 +1,72 @@
...
+
+    for (Map.Entry<String, Double> entry1 : map1.entrySet()) {
+      String float1Key = entry1.getKey();
+      Double float1Value = entry1.getValue();
+
+      if (float1Value > cap) {
+        float1Value = cap;
+      }
+
+      Double float1Quantized = LinearLogQuantizeTransform.quantize(float1Value, bucket);
```
deerzq: perhaps we should allow the user select the quantization method

core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatCrossFloatTransform.java
```diff
@@ -0,0 +1,72 @@
+package com.airbnb.aerosolve.core.transforms;
+
+import com.airbnb.aerosolve.core.FeatureVector;
+import com.airbnb.aerosolve.core.util.Util;
+
+import java.util.Map;
+
+import com.typesafe.config.Config;
+
+public class FloatCrossFloatTransform extends Transform {
```
deerzq: more comments about the transform usage

jq:
:+1:  go ahead and add test + comments, for "user select the quantization method" you can pick either add in this pr or next pr.

core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatCrossFloatTransform.java
christhetree: I'll add this in another PR.

core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatCrossFloatTransform.java
christhetree: Done.

christhetree:
@jq @deerzq Tests and comments added PTAL

core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatCrossFloatTransform.java
deerzq: sounds good

deerzq:
LGTM

jq:
:+1:

PR #120, opened 2016-04-07
update Full rank model default

deerzq:
- rprop performs better than sparse_boost
- don't cache by default

to: @jq @Patrick1860


Patrick1860:
LGTM

PR #122, opened 2016-04-13
[pricing] Add generic pipeline to aerosolve

yolken:
to: @deerzq
cc: @askeys @jq 

This change adds a "generic pipeline" that allows one to train, evaluate, and score an Aerosolve model with minimal code. We've been using this inside Airbnb successfully for the last year. 

The code is the same as the internal version, except for some style cleanups.

Documentation will be added in a future change.


deerzq:
LGTM, some eval utils might be moved to the `Evaluation` later.

PR #124, opened 2016-04-15
Misc. cleanups and performance enhancements to DecisionTreeTrainer

yolken:
to: @deerzq 

Content:
- Clean up style in DecisionTreeTrainer by using enums instead of strings, adding vertical whitespace, etc.
- Do the following performance enhancements (result in ~25% speed gain):
  1. Remove asScala conversions and use Java APIs directly when possible
  2. Partition examples when evaluating sub-trees
  3. Don't do split when size of examples matches minimum size


deerzq:
LGTM!

PR #133, opened 2016-04-19
Update travis setting to fix flaky java installation

saurfang:
This updates the travis setting to use pre-installed jdk8 directly and switch to Linux based build environment. This fixes flaky build issue due to brew install java being flaky. 

It also saves about 5 minutes build time (~30%) per build. 

This also prepares us to move to container based build once the thrift installation issue mentioned below be fixed. (which allows thrift installation and Java packages be cached and faster build startup time)

@jq  @deerzq

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,20 @@
 language: java
-os : osx
+jdk:
+  - oraclejdk8
+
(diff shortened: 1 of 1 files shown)


jq:
👍

deerzq:
LGTM

PR #136, opened 2016-04-20
Move float to string and float transform

christhetree:
to: @jq @deerzq 
cc: @yolken @hectorgon 

This PR adds a transform that takes the floats in fieldName1 and quantizes them into buckets. If the quantized float is less than or equal to a maximum specified bucket value or greater than or equal to a minimum specified bucket value, then the quantized float is stored as a string in a new string feature output specified by stringOutputName. Otherwise, the original, unchanged float is stored in a new float feature output specified by floatOutputName. The input float feature remains unchanged.

This should be useful for working with features that contain the majority of their data in only a few buckets / have a very skewed distribution. The bulk of the data can be pulled out as a categorical string feature with a new float feature being created with the remaining data which can then be processed as desired.


jq:
👍

PR #137, opened 2016-04-20
update and eval method of multi dimension spline and unit test

jq:


deerzq:
LGTM

PR #138, opened 2016-04-20
remove getWeights from Function interface.

jq:

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/function/Linear.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/function/Linear.java
@@ -13,9 +13,7 @@ public class Linear extends AbstractFunction {
   // weights[0] is offset, weights[1] is slope
   public Linear(Linear other) {
     functionForm = FunctionForm.LINEAR;
-    weights = new float[2];
-    weights[0] = other.getWeights()[0];
-    weights[1] = other.getWeights()[1];
+    weights = other.weights.clone();
     minVal = other.getMinVal();
     maxVal = other.getMaxVal();
   }
--- a/core/src/main/java/com/airbnb/aerosolve/core/function/Linear.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/function/Linear.java
@@ -26,6 +26,20 @@ public class Linear extends AbstractFunction {
   }
 
   @Override
+  public Function aggregate(Iterable<Function> functions, float scale, int numBins) {
+    int length = weights.length;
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/airbnb/aerosolve/core/function/Spline.java
```diff
@@ -21,21 +21,23 @@ public Spline(float minVal, float maxVal, float [] weights) {
...
-        float x = diff * t + other.minVal;
-        newWeights[i] = other.evaluate(x);
-      }
+      setupSpline(other.minVal, other.maxVal, other.newBinWeights(newBins));
+    }
+  }
+
+  public float[] newBinWeights(int newBins) {
+    float[] newWeights = new float[newBins];
+    float scale = 1.0f / (newBins - 1.0f);
```
spencerde: `newBins` won't ever be 1 here?

core/src/main/java/com/airbnb/aerosolve/core/function/Spline.java
```diff
@@ -21,21 +21,23 @@ public Spline(float minVal, float maxVal, float [] weights) {
...
-      for (int i = 0; i < newBins; i++) {
-        float t = i * scale;
-        float x = diff * t + other.minVal;
-        newWeights[i] = other.evaluate(x);
-      }
+      setupSpline(other.minVal, other.maxVal, other.newBinWeights(newBins));
+    }
+  }
+
+  public float[] newBinWeights(int newBins) {
```
spencerde: Can you add a comment for what this does, that it generates a new set of weights which mimics the shape of this spline rescaled to a new number of bins?

spencerde:
One style comment, overall LGTM

jq:
@spencerde thanks for review, add comments and name it to weightsByNumBins

core/src/main/java/com/airbnb/aerosolve/core/function/Linear.java
```diff
@@ -28,6 +26,20 @@ public Linear(float minVal, float maxVal, float[] weights) {
   }
 
   @Override
+  public Function aggregate(Iterable<Function> functions, float scale, int numBins) {
```
deerzq: numBins is not used here.

jq: yes, just to keep interface for both linear and spline

deerzq:
We may want to rename numBins to something else, since it's a Spline specific thing. LGTM

PR #140, opened 2016-04-21
change setPrior to String[] so that multi dimension spline can implem…

jq:
…ent it too


deerzq:
Just looking at this change alone, does not feel it's right to do so. What types of input are you expecting to take for the multidimensional spline?

jq:
@deerzq  not very sure right now, maybe I can wait till I know

PR #145, opened 2016-04-22
Rewrite addFunction of AdditiveModel to add new type of function easily

jq:


deerzq:
LGTM, I feel we need to test the recent changes by training on the real data to make sure nothing is broken.

jq:
@deerzq good point, will try that!

PR #148, opened 2016-04-22
Be stricter when loading config file and make JobRunner default main class

saurfang:
We have seen cases where config file was not found and config loader fails silently. It will report a specific key not found instead. This confuses users and make them believe wrong keys have been entered in the config file. 
In addition, we have found the way Spark 1.5+ organize driver classpath make config loader no longer search present working directory anymore. While we can work around it with `--driver-classpath .`, if we consider the purpose of generic pipeline being for user not to build the JAR again, making the config loader look on the filesystem by default seems a better solution. 
We have also improved logging so it is explicit about whether the config file is loaded from filesystem or classpath (or not found at all). 

Finally we have made JobRunner the main class of the training JAR, therefore `--class com.airbnb.aerosolve.training.pipeline.JobRunner` is no longer required.

@jq @deerzq @yolken @coleslaw

### DIFF
--- a/training/build.gradle
+++ b/training/build.gradle
@@ -54,6 +54,10 @@ compileScala {
 
 shadowJar {
   zip64 = true
+
+  manifest {
+    attributes 'Main-Class': 'com.airbnb.aerosolve.training.pipeline.JobRunner'
+  }
 }
 
 test {
--- a/training/src/main/scala/com/airbnb/aerosolve/training/pipeline/JobRunner.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/pipeline/JobRunner.scala
@@ -1,26 +1,31 @@
 package com.airbnb.aerosolve.training.pipeline
 
-import org.apache.spark.{SparkContext, SparkConf}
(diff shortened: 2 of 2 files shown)


jq:
👍

training/build.gradle
```diff
@@ -52,10 +52,18 @@ compileScala {
   scalaCompileOptions.forkOptions.jvmArgs = ['-XX:MaxPermSize=512m']
 }
 
+
```
jq: extra space?

training/build.gradle
```diff
@@ -52,10 +52,18 @@ compileScala {
...
+  manifest {
+    attributes 'Main-Class': 'com.airbnb.aerosolve.training.pipeline.JobRunner'
+  }
 }
 
 test {
   jvmArgs += [ "-XX:MaxPermSize=1024m" ]
 }
+
+
```
jq: extra space?

coleslaw:
Nice work ! 

LGTM

training/build.gradle
saurfang: fixed 😄

PR #150, opened 2016-04-23
Add smooth method to Function

jq:

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/function/FunctionUtil.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/function/FunctionUtil.java
@@ -1,0 +1,77 @@
+package com.airbnb.aerosolve.core.function;
+
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/airbnb/aerosolve/core/function/FunctionUtil.java
```diff
@@ -0,0 +1,77 @@
...
+    int iterations = numCoeff * 4;
+    float[] initial = new float[numCoeff];
+
+    float[] initialStep = new float[numCoeff];
+    Arrays.fill(initialStep, 1.0f);
+    return optimize(1.0 / 512.0, iterations, initial, initialStep,
+        new ImmutablePair<Float, Float>(-10.0f, 10.0f), data);
+  }
+
+  public static float evaluatePolynomial(float[] coeff, float[] data, boolean overwrite) {
```
deerzq: why don't you also delete the existing implementation of these inside SplineTrainer?

jq: how about @Deprecated them? I just update the @Deprecated

deerzq:
LGTM

PR #152, opened 2016-04-25
refactor NDTree so that easy to accept new data types for leaf methods

jq:


core/src/main/java/com/airbnb/aerosolve/core/function/MultiDimensionPoint.java
```diff
@@ -116,7 +121,17 @@ public static float euclideanDistance(float[] x, List<Float> y) {
     return (float) Math.sqrt(sum);
   }
 
-  @Override
+  public static float euclideanDistance(List<Double> x, List<Float> y) {
```
deerzq: we can put this inside util since this is very generic.

jq:
@deerzq Thanks for review, euclideanDistance moved

deerzq:
LGTM

PR #153, opened 2016-04-27
replace type with instanceof

jq:
@deerzq @hectorgon @spencerde


deerzq:
LGTM

bhntr:
Cool.  Looks like this simplified the code a lot!

PR #154, opened 2016-04-27
Add Feature to Example methods to Features and update GenericPipeline…

jq:
… to use it. This make sure model using generic pipeline can do online scoring too.

@deerzq @yolken @hectorgon @spencerde @bhntr

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
@@ -1,14 +1,28 @@
 package com.airbnb.aerosolve.core.features;
 
+import com.airbnb.aerosolve.core.Example;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
```diff
@@ -29,4 +43,114 @@
     Collections.sort(features);
     return features;
   }
+
+  public Example toExample(boolean isMultiClass) {
```
deerzq: I feel this is not really very generic, for example, we are not taking care of dense feature here. Perhaps just add some 'TODO' or comment here so that people know that what this is doing, what this is not doing.

deerzq:
Just one comment, overall LGTM

PR #158, opened 2016-04-30
add cache option for additive and linear trainer

deerzq:
- add cache option to allow faster training on small dataset
- remove the isRanking option in additive model trainer since it's not used

to: @jq @spencerde

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
@@ -38,12 +38,12 @@ object AdditiveModelTrainer {
                                    linfinityCap : Double,
                                    threshold : Double,
                                    lossMod : Int,
-                                   isRanking : Boolean,    // If we have a list based ranking loss
                                    rankMargin : Double,    // The margin for ranking loss
                                    epsilon : Double,       // epsilon used in epsilon-insensitive loss for regression training
                                    initModelPath: String,
                                    linearFeatureFamilies: Array[String],
-                                   priors: Array[String])
+                                   priors: Array[String],
+                                   cache: String)
 
   def train(sc : SparkContext,
             input : RDD[Example],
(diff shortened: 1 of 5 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -299,17 +311,6 @@ object AdditiveModelTrainer {
     return loss
   }
 
-  private def transformExamples(input: RDD[Example],
-                        config: Config,
-                        key: String,
-                        params: AdditiveTrainerParams): RDD[Example] = {
-    if (params.isRanking) {
-      LinearRankerUtils.transformExamples(input, config, key)
```
jq: so we don't do isRanking, what's the difference between transformExamples and makePointwiseFloat, both do combineContextAndItems, but transformExamples unsetContext, what's the role of Context play here?

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -52,7 +52,13 @@ object AdditiveModelTrainer {
     val trainConfig = config.getConfig(key)
     val iterations : Int = trainConfig.getInt("iterations")
     val params = loadTrainingParameters(trainConfig)
-    val transformed = transformExamples(input, config, key, params)
+    val pointwise = LinearRankerUtils.makePointwiseFloat(input, config, key)
+
+    val transformed = params.cache match {
+      case "memory" => pointwise.cache()
```
jq: nice! this should boost the performance!

jq:
Just a knowledge question LGTM

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
deerzq: when `makePointwiseFloat` is used one Example contains only one example, but when `transformExamples` is used one Example can contain multiple examples and it is used for compute ranking loss, if you look at the `getTrainingExamples` in PricingPipeline, you can see how examples are grouped. As for context, I think for both cases we are not using it.

PR #159, opened 2016-04-30
Refactor EvalUtil

deerzq:
- getClassificationAUC in EvalUtil was redundant in that we can use existing AUC implementation in Evaluation to get the same result
- move `scoreExamples` to `PipelineUtil`

to: @jq @spencerde


jq:
👍

PR #163, opened 2016-05-01
Ngram transform

christhetree:
to: @asaluja @jq @deerzq 
cc: @yolken @hectorgon 

This PR contains a new transform for generating Ngrams.

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransform.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransform.java
@@ -15,8 +15,6 @@ import com.typesafe.config.Config;
  * "generateBigrams" specifies whether bigrams should also be generated
  */
 public class DefaultStringTokenizerTransform implements Transform {
-  public static final String BIGRAM_SEPARATOR = " ";
-
   private String fieldName1;
   private String regex;
   private String outputName;
@@ -72,7 +70,7 @@ public class DefaultStringTokenizerTransform implements Transform {
           if (previousToken == null) {
             previousToken = token;
           } else {
-            String bigram = previousToken + BIGRAM_SEPARATOR + token;
+            String bigram = previousToken + NgramTransform.BIGRAM_SEPARATOR + token;
             incrementOutput(bigram, bigramOutput);
             previousToken = token;
           }
@@ -81,7 +79,7 @@ public class DefaultStringTokenizerTransform implements Transform {
     }
   }
 
-  private static void incrementOutput(String key, Map<String, Double> output) {
+  public static void incrementOutput(String key, Map<String, Double> output) {
     if (key == null || output == null) {
       return;
     }
--- a/core/src/main/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransform.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransform.java
@@ -2,11 +2,10 @@ package com.airbnb.aerosolve.core.transforms;
 
 import com.airbnb.aerosolve.core.FeatureVector;
 import com.airbnb.aerosolve.core.util.Util;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransform.java
```diff
@@ -15,8 +15,6 @@
  * "generateBigrams" specifies whether bigrams should also be generated
  */
 public class DefaultStringTokenizerTransform implements Transform {
```
jq: since we have ngram, maybe name this as BigramTransform?

core/src/main/java/com/airbnb/aerosolve/core/transforms/NgramTransform.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Tokenizes strings using a regex and generates ngrams from the tokens
+ * "field1" specifies the key of the feature
+ * "regex" specifies the regex used to tokenize
+ * "n" specifies the size of the ngrams
+ * "min_n" optional parameter, if specified ngrams from min_n (inclusive) to n (inclusive) will
+ *  be generated and placed in the output
+ */
+public class NgramTransform implements Transform {
```
jq: since this is new, we may move it under pkg string, and use it like string.ngram  
this categorizes the Transform from now on.

asaluja:
@christhetree thanks for doing this!

core/src/main/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransform.java
christhetree: I implemented the DefaultStringTokenizerTransform using NgramTransform since it is essentially a 1gram transform with an optional 2gram transform in a separate output. Changing name to bigram doesn't really make sense since that is just an optional functionality of it (main functionality is just to tokenize strings) and should be deprecated in the future since NgramTransform exists now.

core/src/main/java/com/airbnb/aerosolve/core/transforms/NgramTransform.java
christhetree: @jq just to clarify, you mean creating the package ...core.transforms.string right?

jq: yes, and then for the name used in the conf, it will be string.ngram

core/src/main/java/com/airbnb/aerosolve/core/transforms/DefaultStringTokenizerTransform.java
jq: sounds good!

jq:
Hi @christhetree 
  @bhntr  is refactoring the transform, you may want to talk to each other and decide how to merge this in.

jq:
@christhetree  https://github.com/airbnb/aerosolve/tree/bradh_api_and_perf

christhetree:
@jq @bhntr It'll probably be easiest if I merge this after Brad is done refactoring and I can rebase and fix any compatibility problems?

jq:
@christhetree, no reply from @bhntr yet, let me ask him

jq:
@christhetree  you can merge now since @bhntr  postpone the refactoring

core/src/main/java/com/airbnb/aerosolve/core/transforms/NgramTransform.java
christhetree: Done.

christhetree:
@jq sounds good, I'll merge ASAP. Unfortunately, the tests pass on my personal computer, but fail here after adding the transform.string package. I suspect the TransformFactory has to be modified in order support the config file having a transform like "string.ngram" ?

jq:
@christhetree yes, has to change, due to now it assumes flat pkg name. you can choose merge without this

PR #166, opened 2016-05-04
Allow string features use same family name without adding prefix for …

jq:
…each feature.

@deerzq @bhntr @spencerde

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
@@ -4,16 +4,15 @@ import com.airbnb.aerosolve.core.Example;
 import com.airbnb.aerosolve.core.FeatureVector;
 import com.airbnb.aerosolve.core.util.Util;
 import com.google.common.annotations.VisibleForTesting;
-import lombok.experimental.Builder;
-import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 
 import java.lang.reflect.Field;
 import java.util.*;
 
-@Builder @Slf4j
 public class Features {
+  public final static String DEFAULT_STRING_FAMILY = "L";
+  public final static String NO_DEFAULT_STRING_FAMILY = "";
   public final static String LABEL = "LABEL";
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
```diff
@@ -4,16 +4,18 @@
...
-import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 
 import java.lang.reflect.Field;
 import java.util.*;
 
-@Builder @Slf4j
 public class Features {
+  public final static String DEFAULT_STRING_FAMILY = "L";
```
deerzq: I feel we need a more distinguishable name, what if there are other existing feature family called 'L'?

deerzq:
I feel you could just add a prefix "L_" to all string features in our pricing table to avoid having the complication here.

core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
jq: in this case, can use genFeaturesWithCustomerStringFamily

jq:
@deerzq add L_ to all string features will require online scoring to update all the naming too. so add it here to reduce the work on online scoring.

deerzq:
I see.

core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
```diff
@@ -25,6 +27,39 @@
 
   public final String[] names;
   public final Object[] values;
+  private final String defaultStringFamilyName;
+
+  public static Features genDefaultStringFamilyFeatures(String[] names, Object[] values) {
+    return new Features(names, values);
+  }
+
+  public static Features genFeaturesWithCustomerStringFamily(
```
spencerde: sp/Customer/Custom

core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
spencerde: But could this have unintended consequences if people didn't realize this was 'L' and had another family of the same name coincidentally?

PR #167, opened 2016-05-04
AdditiveModelTrainer Refactoring and Documentation

saurfang:
This started as an attempt to refactor AdditiveModelTrainer to make it much faster. However I have come to the conclusion that the trainer implementation is not the bottleneck but rather the inefficient Example class is the culprit for the poor performance.  It roughly takes 5~10x time to load Example data than the actually optimization routine at each iteration.

Before moving to a more efficient representation of feature vectors, I want to merge in the initial refactoring, clean up and documentation I added for the analysis. There should be no performance impact and logic changes for this PR.

@jq @deerzq @yolken 

Let me know if you have any comment or suggestions on the documentation. You will probably find it easier to review commit by commit.

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
@@ -294,8 +294,4 @@ public class AdditiveModel extends AbstractModel {
       }
     }
   }
-
-  public void resample(int newBins) {
-    weights.values().forEach(m -> m.values().forEach(f -> f.resample(newBins)));
-  }
 }
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
@@ -294,4 +294,8 @@ public class AdditiveModel extends AbstractModel {
       }
     }
   }
+
+  public void resample(int newBins) {
+    weights.values().forEach(m -> m.values().forEach(f -> f.resample(newBins)));
+  }
 }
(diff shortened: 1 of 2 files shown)


deerzq:
@bhntr

core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
```diff
@@ -294,4 +294,8 @@ public void updateDense(float gradWithLearningRate,
       }
     }
   }
+
+  public void resample(int newBins) {
```
deerzq: I feel this is only used in multiscale sgd model trainer, it might be better to move this inside trainer so that once the model is trained, the user cannot modify the underlying weight. Just one thought, not sure what's the best design here.

tension7:
I think @bhntr was also working on improving the Example.

core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
jq: one idea is Mutable Model used in Training, and Generate Immutable model from Training. the Mutable model has resample method, while immutable model doesn't

jq:
I think this is not conflict with @bhntr 's Example refactor, and Code is cleaner. can you run the sample's see if performance is same?  others LGTM

bhntr:
I have not touched the training project yet.  My goal was to get the APIs totally solid in core so I would only have to touch the trainers once.  

So, I say A+++ to any code cleanup in training cause it will make my life easier when I get there. I'll hopefully do that tomorrow. I'll check in and make sure this is merged before I start on the training project. Thanks for doing this, Forest!

core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
saurfang: That's a fair point. The motivation of moving to core was to avoid java/scala conversion. I will move it back and use for loop instead.

saurfang:
Thanks all for the reviews and comments. Verified there is no performance change.

PR #170, opened 2016-05-11
[additive model] clean up

deerzq:
to: @jq @yolken

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
@@ -45,8 +45,6 @@ object AdditiveModelTrainer {
                                    linfinityCap: Double,
                                    threshold: Double,
                                    lossMod: Int,
-                                   isRanking: Boolean, // If we have a list based ranking loss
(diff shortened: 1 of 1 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -80,9 +78,9 @@ object AdditiveModelTrainer {
     * During each iteration, we:
     *
     * 1. Sample dataset with subsample (this is analogous to mini-batch sgd?)
-    * 1. Repartition to numBags (this is analogous to ensemble averaging?)
-    * 1. For each bag we run SGD (observation-wise gradient updates)
-    * 1. We then average fitted weights for each feature and return them as updated model
+    * 2. Repartition to numBags (this is analogous to ensemble averaging?)
+    * 3. For each bag we run SGD (observation-wise gradient updates)
+    * 4. We then average fitted weights for each feature and return them as updated model
     *
```
saurfang: Scaladoc requires numbered list use exact same number and it will increment them automatically. Totally weird! https://gist.github.com/VladUreche/8396624#file-gistfile1-txt-L127

jq:
👍

PR #171, opened 2016-05-11
Add NDTree pipeline

jq:
@deerzq @saurfang 
WIP, will continue the save and test.

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
@@ -52,7 +52,7 @@ object AdditiveModelTrainer {
                                    initModelPath: String,
                                    linearFeatureFamilies: java.util.List[String],
                                    priors: Array[String],
-                                   dynamicBuckets: Boolean)
+                                   nDTreeBuildOptions: NDTreeBuildOptions)
 
   def train(sc: SparkContext,
             input: RDD[Example],
@@ -66,7 +66,7 @@ object AdditiveModelTrainer {
     log.info("Training using " + params.loss)
 
     val paramsBC = sc.broadcast(params)
-    var model = modelInitialization(sc, transformed, params, trainConfig)
+    var model = modelInitialization(sc, transformed, params)
     for (i <- 1 to iterations) {
       log.info(s"Iteration $i")
       val modelBC = sc.broadcast(model)
@@ -320,36 +320,32 @@ object AdditiveModelTrainer {
 
   private def modelInitialization(sc: SparkContext,
                                   input: RDD[Example],
-                                  params: AdditiveTrainerParams,
-                                  config: Config): AdditiveModel = {
+                                  params: AdditiveTrainerParams): AdditiveModel = {
     // sample examples to be used for model initialization
     if (params.initModelPath == "") {
       val newModel = new AdditiveModel()
-      initModel(sc, config, params, input, newModel, true)
+      initModel(sc, params, input, newModel, true)
       setPrior(params.priors, newModel)
       newModel
     } else {
       val newModel = TrainingUtils.loadScoreModel(params.initModelPath)
         .get.asInstanceOf[AdditiveModel]
-      initModel(sc, config, params, input, newModel, false)
+      initModel(sc, params, input, newModel, false)
       newModel
     }
   }
 
   // Initializes the model
   private def initModel(sc: SparkContext,
-                        config: Config,
                         params: AdditiveTrainerParams,
                         input: RDD[Example],
                         model: AdditiveModel,
                         overwrite: Boolean) = {
-    if (params.dynamicBuckets) {
+    if (params.nDTreeBuildOptions != null) {
       val linearFeatureFamilies = params.linearFeatureFamilies
-      val options = NDTreeBuildOptions(
-        maxTreeDepth = config.getInt("max_tree_depth"),
-        minLeafCount = config.getInt("min_leaf_count"))
       val result: Array[((String, String), Any)] = NDTreePipeline.getFeatures(
-        sc, input, params.minCount, params.subsample, linearFeatureFamilies, options)
+        sc, input, params.minCount, params.subsample,
+        linearFeatureFamilies, params.nDTreeBuildOptions)
       for (((family, name), feature) <- result) {
         // save to disk.
         if (feature.isInstanceOf[FeatureStats]) {
@@ -474,7 +470,15 @@ object AdditiveModelTrainer {
     val multiscale: Array[Int] = Try(
       config.getIntList("multiscale").asScala.map(x => x.toInt).toArray)
       .getOrElse(Array[Int]())
-    val dynamicBuckets = Try(config.getBoolean("dynamic_buckets")).getOrElse(false)
+    val dynamicBucketsConfig = Try(Some(config.getConfig("dynamic_buckets"))).getOrElse(None)
+    val options = if (dynamicBucketsConfig.nonEmpty) {
+      val cfg = dynamicBucketsConfig.get
+      NDTreeBuildOptions(
+        maxTreeDepth = cfg.getInt("max_tree_depth"),
+        minLeafCount = cfg.getInt("min_leaf_count"))
+    } else {
+      null
+    }
 
     AdditiveTrainerParams(
       numBins,
@@ -496,7 +500,7 @@ object AdditiveModelTrainer {
       initModelPath,
       linearFeatureFamilies,
       priors,
-      dynamicBuckets)
+      options)
   }
 
   def trainAndSaveToFile(sc: SparkContext,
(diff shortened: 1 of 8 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
```diff
@@ -0,0 +1,110 @@
...
+      val denseFeatures = Util.safeMap(featureVector.denseFeatures).asScala
+      denseFeatures.foreach(feature => {
+        val name = feature._1
+        val values:ArrayBuffer[Array[Double]] =
+          map.getOrElseUpdate((AdditiveModel.DENSE_FAMILY, name),
+            ArrayBuffer[Array[Double]]())
+        values += feature._2.asScala.toArray.map(_.doubleValue)
+      })
+    }
+  }
```
saurfang: I remember seeing these flatten function somewhere already?

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/JobRunner.scala
```diff
@@ -72,6 +72,7 @@ object JobRunner extends Logging {
             .scoreTableRun(sc, config)
           case "GenericParamSearch" => GenericPipeline
             .paramSearch(sc, config)
+          case "BuildFeatureMap" => NDTreePipeline.buildFeatureMapRun(sc, config)
```
deerzq: "BuildFeatureMap" sounds very generic, rename as "NDTreeBuildFeatureMap" ?

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
```diff
@@ -0,0 +1,110 @@
...
+      a.++=(b)
+    }).filter(x => x._2.size >= minCount)
+      // build tree
+      .map(x => ((x._1._1, x._1._2), NDTree(options.value, x._2.toArray).nodes))
+      .collect
+    for (((family, name), features) <- tree) {
+      // save to disk.
+      log.info(s"${family}, ${name}: ${features.mkString(" ")}")
+
+    }
```
deerzq: so this is not returning or save anything?

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
```diff
@@ -0,0 +1,110 @@
...
+      .getOrElse[java.util.List[String]](List.empty.asJava)
+
+    val linearFeatureFamiliesBC = sc.broadcast(linearFeatureFamilies)
+    val options = sc.broadcast(NDTreeBuildOptions(
+      maxTreeDepth = cfg.getInt("max_tree_depth"),
+      minLeafCount = cfg.getInt("min_leaf_count")))
+
+    log.info("Training data: ${inputPattern}")
+    val input = GenericPipeline.getExamples(sc, inputPattern).sample(true, sample)
+    val tree: Array[((String, String), Array[NDTreeNode])] = input.mapPartitions(partition => {
```
deerzq: return `Array[((String, String), NDTreeModel)]`?

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
jq: the flattern combine string and float together, I handle them differently

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
jq: updated with return method and print out for debug

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -314,38 +318,63 @@ object AdditiveModelTrainer {
...
                         params: AdditiveTrainerParams,
-                        examples: RDD[Example],
+                        input: RDD[Example],
                         model: AdditiveModel,
                         overwrite: Boolean) = {
+    if (params.dynamicBuckets) {
+      val linearFeatureFamilies = params.linearFeatureFamilies
+      val options = NDTreeBuildOptions(
+        maxTreeDepth = config.getInt("max_tree_depth"),
+        minLeafCount = config.getInt("min_leaf_count"))
```
deerzq: instead of passing in config in `initModel`, how about including these parameters in `AdditiveTrainerParams`

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -314,38 +318,63 @@ object AdditiveModelTrainer {
...
+    } else {
+      initModel(params, input, model, overwrite)
+    }
+  }
+
+  // init spline and linear
+  private def initModel(params: AdditiveTrainerParams,
+                        input: RDD[Example],
+                        model: AdditiveModel,
+                        overwrite: Boolean)= {
```
deerzq: I don't see where this is used

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
```diff
@@ -0,0 +1,194 @@
...
+      sc, input, minCount, sample, linearFeatureFamilies, options)
+    result
+  }
+
+  def getFeatures(sc: SparkContext, input: RDD[Example], minCount: Int, sample: Double,
+                  linearFeatureFamilies: util.List[String],
+                  options: NDTreeBuildOptions): Array[((String, String), Any)] = {
+    val linearFeatureFamiliesBC = sc.broadcast(linearFeatureFamilies)
+    val optionsBC = sc.broadcast(options)
+    val tree: Array[((String, String), Any)] = input.sample(true, sample).mapPartitions(partition => {
```
deerzq: the returning type `Any` seems too arbitrary, can we have an actual type here?

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
```diff
@@ -0,0 +1,194 @@
...
+      }
+    }).map(x => {
+      val a = x._2
+      if (a.isInstanceOf[ArrayBuffer[Array[Double]]]) {
+        // build tree
+        ((x._1._1, x._1._2), NDTree(optionsBC.value,
+          x._2.asInstanceOf[ArrayBuffer[Array[Double]]].toArray).model)
+      } else {
+        x
+      }
```
deerzq: I feel there are too many `asInstanceOf` and `Any` in the above code, there should be better way to write

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
saurfang: AFAIK, there is no union type in Scala out of the box. You can use `Either` type to make type more precise. Another idea is since feature family has low cardinality (you are collecting at the end) we could really aggregate/reduce with a Map (i.e. not partition by key at all).

jq: 1) “use Either type” ？ what about the other type?
2) to compute NDTree, I need all numbers together, the other type I guess it is possible to do aggregate with a map, but then you still out put with different type...

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
jq: only NDTree use it, maybe I should put dynamicBuckets as a new section dynamic_buckets {} and put them inside

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
jq: Thanks for suggestion!, I will try Either Type

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
```diff
@@ -0,0 +1,203 @@
...
+    result
+  }
+
+  def getFeatures(sc: SparkContext, input: RDD[Example], minCount: Int, sample: Double,
+                  linearFeatureFamilies: util.List[String],
+                  options: NDTreeBuildOptions):
+                  Array[((String, String), Either[NDTreeModel, FeatureStats])] = {
+    val linearFeatureFamiliesBC = sc.broadcast(linearFeatureFamilies)
+    val optionsBC = sc.broadcast(options)
+    val tree: Array[((String, String), Either[NDTreeModel, FeatureStats])] = input.sample(true, sample).mapPartitions(partition => {
```
deerzq: this line is too long

deerzq:
some nits, LGTM

PR #172, opened 2016-05-12
Performance improvements from Forest Fang, by not doing a full transf…

jq:
…orm on all samples

@saurfang @deerzq

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
@@ -1,4 +1,4 @@
-package com.airbnb.aerosolve.training
+package com.airbnb.marketplace.pricing
 
 import com.airbnb.aerosolve.core._
 import com.airbnb.aerosolve.core.function.{Function, Linear, MultiDimensionSpline, Spline}
@@ -12,6 +12,7 @@ import org.apache.spark.SparkContext
 import org.apache.spark.broadcast.Broadcast
 import org.apache.spark.rdd.RDD
 import org.slf4j.{Logger, LoggerFactory}
+import com.airbnb.aerosolve.training._
 
 import scala.collection.JavaConversions._
 import scala.collection.JavaConverters._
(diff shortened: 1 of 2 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -1,4 +1,4 @@
-package com.airbnb.aerosolve.training
+package com.airbnb.marketplace.pricing
```
deerzq: pricing?

jq: nice catch, updated!

saurfang:
Added more documentation around the motivation and caveat of early sampling. Added wrapper functions for backwards compatibility. Removed duplicated sampling in model initialization. 
@jq Did you mean to sample *with* replacement in NDTree or was it an oversight? (This changes it to sample without replacement)

jq:
@saurfang I pass in true since it doesn't need to be remember sampled for NDTree, a slightly better performance?

jq:
👍  let's get it in and try

deerzq:
LGTM

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -53,15 +53,19 @@ object AdditiveModelTrainer {
                                    linearFeatureFamilies: java.util.List[String],
                                    priors: Array[String],
                                    nDTreeBuildOptions: NDTreeBuildOptions)
+  def train(sc: SparkContext, input: RDD[Example], config: Config, key: String): AdditiveModel =
+    train(sc, (frac: Double) => input.sample(false, frac), config, key)
```
yinlou: Why do we need this? I thought we subsampled before calling this train?

saurfang: For future reference, we want to sample during each iteration and initialization. We prefer not to double sample during model training.

PR #173, opened 2016-05-13
Make Function and AdditiveModel cloneable

saurfang:
Currently AdditiveModel support multiscale sampling and does so by assigning each partition one scale to train model on. Each partition receives the (mutable) model spec/weights via broadcasted variable, which each JVM has one and only one copy. This becomes a problem when each executor has multi-cores and tasks on the same executor work on different partitions/scales. This create a race-condition between executors and leads to ArrayIndexOutOfBound exception.

Even when multiscaling is off, task runner on the same executor will still overwrite each other's weight update and race condition simply gets masked.

We work around this issue by making Function and AdditiveModel cloneable such that a deep copy is made for each partition. In the long run, we should revisit the liberal use of mutable data structure in Aerosolve and possibly adopt a different weights updating scheme.

@jq @deerzq @yolken @timyitong

p.s. Never done Java cloneable before. Please let me know if you have a better idea. (I considered copy constructor and builder pattern but couldn't come up a way to make inheritance work.)

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/function/Function.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/function/Function.java
@@ -5,7 +5,7 @@ import com.airbnb.aerosolve.core.ModelRecord;
 import java.io.Serializable;
 import java.util.List;
 
-public interface Function extends Serializable {
+public interface Function extends Serializable, Cloneable {
   // TODO rename numBins to something else, since it's a Spline specific thing
   Function aggregate(Iterable<Function> functions, float scale, int numBins);
 
@@ -27,4 +27,6 @@ public interface Function extends Serializable {
   void resample(int newBins);
 
   void smooth(double tolerance);
+
+  Function clone() throws CloneNotSupportedException;
 }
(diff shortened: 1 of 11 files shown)


deerzq:
nice catch! LGTM

core/src/main/java/com/airbnb/aerosolve/core/function/Function.java
```diff
@@ -5,7 +5,7 @@
 import java.io.Serializable;
 import java.util.List;
 
-public interface Function extends Serializable {
+public interface Function extends Serializable, Cloneable {
```
jq: usually copy constructor is favor over clone, how about  Function deepCopy() in the interface

core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
```diff
@@ -23,7 +23,7 @@
 // A generalized additive model with a parametric function per feature.
 // See http://en.wikipedia.org/wiki/Generalized_additive_model
 @Slf4j
-public class AdditiveModel extends AbstractModel {
+public class AdditiveModel extends AbstractModel implements Cloneable {
```
jq: Add it to Model as Model deepCopy() maybe too big at this time, how bout  public AdditiveModel deepCopy() ?

saurfang:
What's the benefit of using `deepCopy` instead of `clone`? The `clone` is `protected` scoped by default and is not accessible unless we override. Is it just to clarify the fact we are deep copying vs. shallow copying?

jq:
@saurfang it is just clarify the fact we are deep copying, I am ok with check this in and refactor later, since the ideal solution is ModelTrainer can be shared. and this deep copy is part of Model operation

PR #175, opened 2016-05-15
allow StringCrossFloatTransform to cross individual features

jq:
@deerzq @saurfang

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
@@ -120,6 +120,19 @@ public class Features {
     }
   }
 
+  // string feature is concatenated by : the prefix before : is feature name
+  // RAW feature has no : so just return the RAW
+  // this is used in StringCrossFloatTransform so that
+  // we can cross Raw feature as well as other string features
+  public static String getStringFeatureName(String feature) {
+    String[] tokens =  feature.split(":");
+    if (tokens.length == 1) {
+      return RAW;
+    } else {
+      return tokens[0];
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/airbnb/aerosolve/core/transforms/StringCrossFloatTransform.java
```diff
@@ -1,23 +1,37 @@
...
-import java.util.HashMap;
-import java.util.Set;
-import java.util.Map;
 
+import java.util.*;
+
+/*
+  if no prefix/keys provided, then cross whole family.
+  prefix used in string, and keys used in float.
+  to cross RAW features, use Features.RAW
```
deerzq: please add more details about what is key, what is prefix, without actually looking at the code, it's not clear that prefix is for string family and key is for float feature. and it's also not clear what exactly is prefix, people may think it means the prefix of a feature name, but it's actually feature name.

jq:
@deerzq thanks for review, updated!

core/src/main/java/com/airbnb/aerosolve/core/features/Features.java
```diff
@@ -120,6 +120,19 @@ static void addStringFeature(
     }
   }
 
+  // string feature is concatenated by : the prefix before : is feature name
+  // RAW feature has no : so just return the RAW
+  // this is used in StringCrossFloatTransform so that
+  // we can cross Raw feature as well as other string features
+  public static String getStringFeatureName(String feature) {
+    String[] tokens =  feature.split(":");
```
deerzq: make ":" a constant?

deerzq:
one more comment, LGTM

PR #176, opened 2016-05-15
Add Point function, so that string feature can use it instead of line…

jq:
…ar function

@deerzq @saurfang

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/function/Linear.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/function/Linear.java
@@ -98,6 +98,7 @@ public class Linear extends AbstractFunction {
 
   @Override
   public float LInfinityNorm() {
+    // return max contribution
     float f0 = weights[0];
     float f1 = weights[0] + weights[1];
     return Math.max(Math.abs(f0), Math.abs(f1));
--- a/core/src/main/java/com/airbnb/aerosolve/core/function/Linear.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/function/Linear.java
@@ -98,7 +98,7 @@ public class Linear extends AbstractFunction {
 
   @Override
   public float LInfinityNorm() {
-    // return max contribution
+    // return max absolute contribution
     float f0 = weights[0];
     float f1 = weights[0] + weights[1];
     return Math.max(Math.abs(f0), Math.abs(f1));
--- a/core/src/main/java/com/airbnb/aerosolve/core/function/MultiDimensionSpline.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/function/MultiDimensionSpline.java
@@ -247,10 +247,9 @@ public class MultiDimensionSpline implements Function {
     }
   }
 
-  // TODO: this is a shallow copy. implement deep copy if need immutability
   @Override
   public MultiDimensionSpline clone() throws CloneNotSupportedException {
-    return (MultiDimensionSpline) super.clone();
+    return new MultiDimensionSpline(this.ndTreeModel);
   }
 
   private boolean canDoSmooth() {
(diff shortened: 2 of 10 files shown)


core/src/main/java/com/airbnb/aerosolve/core/function/Linear.java
```diff
@@ -98,6 +98,7 @@ public void LInfinityCap(float cap) {
 
   @Override
   public float LInfinityNorm() {
+    // return max contribution
```
deerzq: max absolute contribution

core/src/main/java/com/airbnb/aerosolve/core/function/MultiDimensionSpline.java
```diff
@@ -247,10 +247,9 @@ public void smooth(double tolerance) {
     }
   }
 
-  // TODO: this is a shallow copy. implement deep copy if need immutability
   @Override
   public MultiDimensionSpline clone() throws CloneNotSupportedException {
-    return (MultiDimensionSpline) super.clone();
+    return new MultiDimensionSpline(this.ndTreeModel);
```
deerzq: why this change?

core/src/main/java/com/airbnb/aerosolve/core/function/Point.java
```diff
@@ -0,0 +1,93 @@
...
+    this.weight = weight;
+  }
+
+  public Point(ModelRecord record) {
+    List<Double> weightVec = record.getWeightVector();
+    weight = weightVec.get(0).floatValue();
+  }
+
+  @Override
+  public Function aggregate(Iterable<Function> functions, float scale, int numBins) {
```
deerzq: not for this PR, but `numBins` is not used, I feel we may want to make `numBins` available only to Spline related features, otherwise, whenever we add new function type, we have to add this.

core/src/main/java/com/airbnb/aerosolve/core/function/Point.java
```diff
@@ -0,0 +1,93 @@
...
+    float aggWeight = 0;
+    for (Function fun: functions) {
+      Point point = (Point) fun;
+      aggWeight += scale * point.weight;
+    }
+    return new Point(aggWeight);
+  }
+
+  @Override
+  public float evaluate(float... x) {
```
deerzq: for string feature, we assume `x=1`, probably we should return `weight * x` so that using `Point` to represent features in AdditiveModel would be make this additive model equivalent to conventional linear model.

deerzq: By doing this, `Point` can also be used for other float features, not just string features

core/src/main/java/com/airbnb/aerosolve/core/function/MultiDimensionSpline.java
jq: this is the correct way to clone it

core/src/main/java/com/airbnb/aerosolve/core/function/Point.java
jq: this is to hide the implementation details

core/src/main/java/com/airbnb/aerosolve/core/function/Point.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Override
+  public void LInfinityCap(float cap) {
+    if (cap <= 0.0f) return;
+    weight = Math.min(cap, weight);
+  }
+
+  @Override
+  public float LInfinityNorm() {
+    return weight;
```
deerzq: `Math.abs(weight)`, the definition of `LInfinityNorm` won't be consistent with the definition in `Linear` and `Spline` unless we also add `minVal` and `maxVal` as member variables, then the norm would be `Math.max(Math.abs(weight * minVal), Math.abs(weight * maxVal))`. But for the current string feature use case, `Math.abs(weight)` would be good. Perhaps add a `TODO` here.

deerzq:
LGTM

PR #177, opened 2016-05-16
Added downsample and classweights

yinlou:
Added 'downsample' to config, downsample samples a portion of data given a label (stratified sampling).

Added 'classweights' to config. This is used for up-sampling by increasing
the weight on minority class.

Added unit test for downsampling.

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
@@ -52,7 +52,9 @@ object AdditiveModelTrainer {
                                    initModelPath: String,
                                    linearFeatureFamilies: java.util.List[String],
                                    priors: Array[String],
-                                   nDTreeBuildOptions: NDTreeBuildOptions)
+                                   nDTreeBuildOptions: NDTreeBuildOptions,
+                                   classWeights: Map[Int, Float])
+
   def train(sc: SparkContext, input: RDD[Example], config: Config, key: String): AdditiveModel =
     train(sc, (frac: Double) => input.sample(false, frac), config, key)
 
@@ -63,6 +65,7 @@ object AdditiveModelTrainer {
     val trainConfig = config.getConfig(key)
     val iterations: Int = trainConfig.getInt("iterations")
     val params = loadTrainingParameters(trainConfig)
+
     // sample before we transform as it can be very expensive especially for crossing
     // NB: this assumes we don't add/remove observations during transformation
     val transformed = (frac: Double) => transformExamples(input(frac), config, key, params)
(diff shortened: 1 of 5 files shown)


deerzq:
Unit test is failing: https://travis-ci.org/airbnb/aerosolve/builds/130624410#L1583

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -255,7 +259,7 @@ object AdditiveModelTrainer {
     val expCorr = scala.math.exp(corr)
     val loss = scala.math.log(1.0 + 1.0 / expCorr)
     val grad = -label / (1.0 + expCorr)
-    val gradWithLearningRate = grad.toFloat * params.learningRate.toFloat
+    val gradWithLearningRate = grad.toFloat * params.learningRate.toFloat * classWeights(label.toInt)
```
deerzq: additive model also has `regression` loss, the label for regression may not be integer. The customized sampling should  only be applied to classification case and if not specified not affecting the regression learning

yinlou: The classWeights only applies on classification so I only edited updateHinge and updateLogistic. Is 'hinge' also considered as regression?

deerzq: oh, never mind, then you may also want to add this to `updateLogistic`?

yinlou: I think this line is under `updateLogistic` :), and you can find another line below for `updateHinge`

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
```diff
@@ -179,6 +203,16 @@ object TrainingUtils {
     }
   }
 
+  def getLabel(example: Example, loss: String, rankKey: String, threshold: Double): Int = {
+    val fv = example.getExample.get(0)
+    val label: Int = if (loss == "regression") {
+      TrainingUtils.getLabel(fv, rankKey).toInt
```
deerzq: why `toInt`?

yinlou: Since we know this is a classification label, and I use `int` to create the maps (classweights and downsample), that will make the context clearer and can usually prevent potential bugs.

deerzq: but the loss is regression, the label should be double. If this is returning classification label, perhaps we should not allow the loss to be 'regression'

yinlou: Ah I see your point. Updated the code!

deerzq:
Great work! LGTM after fixing build failure

PR #184, opened 2016-05-20
Fix MultiDimension spline save feature name and min/max

jq:
@deerzq @saurfang


saurfang:
hmm. LGTM I guess :)

deerzq:
LGTM

PR #185, opened 2016-05-24
fix NDTreeTraining Build Tree.

jq:
NDTreeTraining build wrong tree 
1. not obey min count option, as long as left or right < min count, it stop
2. it may not create bucket evenly if median happens to be the min element.

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
@@ -2,6 +2,7 @@ package com.airbnb.aerosolve.training
 
 import com.airbnb.aerosolve.core.NDTreeNode
 import com.airbnb.aerosolve.core.models.NDTreeModel
+import org.slf4j.LoggerFactory
 
 import scala.collection.mutable.ArrayBuffer
 import scala.collection.JavaConverters._
@@ -17,6 +18,7 @@ object SplitType extends Enumeration {
  * Companion object to NDTree.
  */
 object NDTree {
+  val log = LoggerFactory.getLogger("NDTree")
   val splitUsingSurfaceAreaMaxCount = 10
 
   val defaultOneDimensionalSplitType = SplitType.Median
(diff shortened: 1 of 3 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
```diff
@@ -158,26 +164,53 @@ object NDTree {
     })
   }
 
-  private def getSplitUsingMedian(
-      points: Array[Array[Double]],
-      indices: Array[Int],
-      axisIndex: Int): Split = {
-    val splitValue = getMedian(points, indices, axisIndex)
+  def nextGreaterElement(
```
deerzq: private?

training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
```diff
@@ -158,26 +164,53 @@ object NDTree {
...
+                            axisIndex: Int,
+                            splitValue: Double):(Array[Int], Array[Int]) ={
     val leftIndices = indices.filter((index: Int) => {
       points(index)(axisIndex) < splitValue
     })
     val rightIndices = indices.filter((index: Int) => {
       points(index)(axisIndex) >= splitValue
     })
+    (leftIndices, rightIndices)
+  }
```
saurfang: You should be able to use this instead
```scala
val (leftIndices, rightIndices) = indices.partition(index => points(index)(axisIndex) < splitValue)
```

deerzq:
LGTM

jq:
@deerzq @saurfang Thanks for review! all updated!

training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
```diff
@@ -186,13 +206,13 @@ object NDTree {
       axisIndex: Int): Double = {
     // TODO (christhetree): use median of medians algorithm (quick select) for O(n)
     // performance instead of O(nln(n)) performance
```
saurfang: breeze has quick select implementation. we can make this change if this is a hot spot in NDTree.

saurfang:
LGTM

PR #186, opened 2016-05-24
add MAP metric

Patrick1860:
This PR corrects the precision @k metrics for multiclass
It should be defined as the proportion of relevant/correct results in top k. The Mean Average Precision(MAP) metrics is also added.

The current definition of precision @k is still useful but needs to be renamed to something else. Here I rename it to accuracy@k

Test passed

@deerzq @jq @hectorgon

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
@@ -117,12 +117,19 @@ object Evaluation {
         }
         metrics.append((prefix + "ALL_PAIRS_HINGE_LOSS", (hingeLoss, 1.0)))
         var inTopK = false
+        var nAccurateLabel = 0.0
+        var sumPrecision = 0.0
         for (i <- 0 until sorted.size) {
           if (rec.labels.containsKey(sorted(i)._1)) {
             inTopK = true
+            nAccurateLabel += 1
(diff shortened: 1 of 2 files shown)


Patrick1860:
FYI @asaluja

asaluja:
@Patrick1860 I know we spoke about this a few months ago, thanks for making this change!

training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
```diff
@@ -117,12 +117,19 @@ object Evaluation {
         }
         metrics.append((prefix + "ALL_PAIRS_HINGE_LOSS", (hingeLoss, 1.0)))
         var inTopK = false
+        var nAccurateLabel = 0.0
+        var sumPrecision = 0.0
         for (i <- 0 until sorted.size) {
```
jq: for (i <- 1 to sorted.size)  then you don't need i+1

hectorgon:
+1

training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
Patrick1860: will need i to do sorted(i)._1 , I think even change it to sorted(i-1)._1 will still need to change the for loop to (i <- 1 until sorted.size + 1), so maybe just keep the a little verbose way

PR #187, opened 2016-05-24
Add checkpoint before additive model training so no rerun coalesce

jq:
@deerzq @saurfang

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
@@ -62,6 +62,7 @@ object AdditiveModelTrainer {
                                    threshold: Double,
                                    epsilon: Double, // epsilon used in epsilon-insensitive loss for regression training
                                    init: InitParams,
+                                   shuffle: Boolean,
                                    classWeights: Map[Int, Float],
                                    checkPointDir: String)
 
@@ -133,7 +134,7 @@ object AdditiveModelTrainer {
     val model = modelBC.value
     val params = sgdParams.paramsBC.value
     val data = input(params.subsample)
-      .coalesce(params.numBags, true)
+      .coalesce(params.numBags, params.shuffle)
 
     if (!params.checkPointDir.isEmpty) {
       data.sparkContext.setCheckpointDir(params.checkPointDir)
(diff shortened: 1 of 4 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -124,15 +126,21 @@ object AdditiveModelTrainer {
...
-
-    input(params.subsample)
+    val data = input(params.subsample)
       .coalesce(params.numBags, true)
-      .mapPartitionsWithIndex((index, partition) => sgdPartition(index, partition, modelBC, sgdParams))
+
+    if (!params.checkPointDir.isEmpty) {
+      sc.setCheckpointDir(params.checkPointDir)
+      data.checkpoint()
+    }
```
saurfang: `sc` can be retrieved directly via `data.context`

saurfang:
I am not sure if this will necessarily help you because transformed examples are one order of magnitude larger so they are slower to write and read in the event of failure, on the other hand checkpointing does potentially avoid shuffle the step in the event of failure. It is hard to say. 
If we see SGD fails often especially due to OOM, we can look into mini-batch SGD instead where gradient is updated per epoch so it can be computed in parallel.

jq:
@saurfang Thanks for review, updated!

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -56,11 +61,10 @@ object AdditiveModelTrainer {
...
                                    epsilon: Double, // epsilon used in epsilon-insensitive loss for regression training
-                                   initModelPath: String,
-                                   linearFeatureFamilies: java.util.List[String],
-                                   priors: Array[String],
-                                   nDTreePipelineParams: NDTreePipelineParams,
-                                   classWeights: Map[Int, Float])
+                                   init: InitParams,
+                                   shuffle: Boolean,
+                                   classWeights: Map[Int, Float],
+                                   checkPointDir: String)
```
deerzq: we are adding more and more parameters here. you may want to write some comments to describe what are these

deerzq:
Please add detailed documentation, then LGTM

PR #188, opened 2016-05-25
use spline if NDTree's max_std_div_by_min_leaf_nodes_count is low

jq:
add new setting max_std_div_by_min_leaf_nodes_count
if NDTree's sdt/min > max_std_div_by_min_leaf_nodes_count then replace it with spline to simplify model.

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/NDTreeModel.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/NDTreeModel.java
@@ -5,6 +5,7 @@ import com.airbnb.aerosolve.core.util.Util;
 import com.google.common.base.Optional;
 import lombok.Getter;
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.math3.stat.descriptive.DescriptiveStatistics;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -40,6 +41,32 @@ public class NDTreeModel implements Serializable {
 
   public NDTreeModel(List<NDTreeNode> nodeList) {
     this(nodeList.toArray(new NDTreeNode[nodeList.size()]));
+  }
+
+  public double stdDivByMinFromLeafNodesCount() {
+    // Get a DescriptiveStatistics instance
+    DescriptiveStatistics stats = new DescriptiveStatistics();
+
+    for (NDTreeNode n : nodes) {
+      if (isLeaf(n)) {
+        stats.addValue(n.count);
+      }
+    }
+    double std = stats.getStandardDeviation();
+    return stats.getMin()/std;
+  }
+
+  /*
+    NDTreeModel is close to evenly distributed and it is 1 dimension
+    So we can replace it by Spline
+   */
+  public boolean canReplaceBySpline(double maxStdDivByMinFromLeafNodesCount) {
+    return dimension == 1 &&
+        maxStdDivByMinFromLeafNodesCount < stdDivByMinFromLeafNodesCount();
+  }
+
+  public static boolean isLeaf(NDTreeNode node) {
+    return node.getAxisIndex() == NDTreeModel.LEAF;
   }
 
   public int leaf(float ... coordinates) {
@@ -92,7 +119,7 @@ public class NDTreeModel implements Serializable {
     }
     return idx;
   }
-  
+
   public static Optional<NDTreeModel> readFromGzippedStream(InputStream inputStream) {
     List<NDTreeNode> nodes = Util.readFromGzippedStream(NDTreeNode.class, inputStream);
     if (!nodes.isEmpty()) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/airbnb/aerosolve/core/models/NDTreeModel.java
```diff
@@ -42,6 +43,32 @@ public NDTreeModel(List<NDTreeNode> nodeList) {
...
+  public double stdDivByMinLeafNodesCount() {
+    // Get a DescriptiveStatistics instance
+    DescriptiveStatistics stats = new DescriptiveStatistics();
+
+    for (NDTreeNode n : nodes) {
+      if (isLeaf(n)) {
+        stats.addValue(n.count);
+      }
+    }
+    double std = stats.getStandardDeviation();
```
deerzq: standard deviation is easy to implement. If std is the only thing we need from the math3 library, perhaps we can just implement ourselves without adding the dependency on the external package.

core/src/main/java/com/airbnb/aerosolve/core/models/NDTreeModel.java
```diff
@@ -42,6 +43,32 @@ public NDTreeModel(List<NDTreeNode> nodeList) {
     this(nodeList.toArray(new NDTreeNode[nodeList.size()]));
   }
 
+  public double stdDivByMinLeafNodesCount() {
```
deerzq: `minLeafNodesDivByStd`?

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -402,11 +402,16 @@ object AdditiveModelTrainer {
...
-            if (stats.min == stats.max) {
-              model.addFunction(family, name, new Point(stats.min.toFloat), overwrite)
+            if (stats.spline) {
+              val spline = new Spline(stats.min.toFloat, stats.max.toFloat, additiveTrainerParams.numBins)
+              model.addFunction(family, name, spline, overwrite)
             } else {
-              model.addFunction(family, name,
-                new Linear(stats.min.toFloat, stats.max.toFloat), overwrite)
+              if (stats.min == stats.max) {
+                model.addFunction(family, name, new Point(stats.min.toFloat), overwrite)
```
saurfang: style. can be `else if {} else {}` instead of `else { if {} else {} }`

core/src/main/java/com/airbnb/aerosolve/core/models/NDTreeModel.java
jq: agree, I have another TODO to use math3, so maybe let's use it from now?

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -402,11 +402,16 @@ object AdditiveModelTrainer {
             model.addFunction(family, name, new  MultiDimensionSpline(ndTreeModel), overwrite)
           }
           case Right(stats) => {
-            if (stats.min == stats.max) {
-              model.addFunction(family, name, new Point(stats.min.toFloat), overwrite)
+            if (stats.spline) {
```
deerzq: what is `stats.spline`?

jq:
@deerzq @saurfang Thanks for review! how about I address both comments in another PR, I have another set of commits in same branch since it touches same files, I just try to avoid the merge so put it together.

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
jq: this is to indicated what function to use

core/src/main/java/com/airbnb/aerosolve/core/models/NDTreeModel.java
deerzq: any reference for using this metric to evaluate uniformity?

jq: it is std/min,  higher means more diff
no reference, if diff is not too much, ndtree is similar to spline.

jq:
@deerzq @saurfang  Thanks for review all comments answered or addressed.

core/src/main/java/com/airbnb/aerosolve/core/models/NDTreeModel.java
deerzq: it is min/std or std/min?

core/src/main/java/com/airbnb/aerosolve/core/models/NDTreeModel.java
```diff
@@ -42,6 +43,32 @@ public NDTreeModel(List<NDTreeNode> nodeList) {
...
+    // Get a DescriptiveStatistics instance
+    DescriptiveStatistics stats = new DescriptiveStatistics();
+
+    for (NDTreeNode n : nodes) {
+      if (isLeaf(n)) {
+        stats.addValue(n.count);
+      }
+    }
+    double std = stats.getStandardDeviation();
+    return stats.getMin()/std;
```
deerzq: is it possible to have std = 0 in the case where all node numbers are the same

deerzq:
LGTM

PR #191, opened 2016-05-25
smoothByTolerancePercentage

jq:
add new smoothByTolerancePercentage method to Function interface and use it in additive model

@deerzq @saurfang

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/function/Function.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/function/Function.java
@@ -28,9 +28,7 @@ public interface Function extends Serializable, Cloneable {
 
   // for function not support smooth, just return 0
   // for function support smooth, return errAndCoeff
-  double smooth(double tolerance);
-  // for function support smooth, return errAndCoeff percentage
-  double smoothByTolerancePercentage(double tolerancePercentage);
+  double smooth(double tolerance, boolean toleranceIsPercentage);
 
   Function clone() throws CloneNotSupportedException;
 }
--- a/core/src/main/java/com/airbnb/aerosolve/core/function/Function.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/function/Function.java
@@ -28,5 +28,7 @@ public interface Function extends Serializable, Cloneable {
 
   void smooth(double tolerance);
 
+  void smoothByTolerancePercentage(double tolerancePercentage);
+
   Function clone() throws CloneNotSupportedException;
 }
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/airbnb/aerosolve/core/function/Function.java
```diff
@@ -26,7 +26,11 @@
 
   void resample(int newBins);
 
-  void smooth(double tolerance);
+  // for function not support smooth, just return 0
+  // for function support smooth, return errAndCoeff
+  float smooth(double tolerance);
+  // for function support smooth, return errAndCoeff percentage
+  float smoothByTolerancePercentage(double tolerancePercentage);
```
saurfang: Do these need to return float or double? Slightly odd when the tolerance is in double.

jq: agree, updated!

core/src/main/java/com/airbnb/aerosolve/core/function/FunctionUtil.java
```diff
@@ -90,16 +92,32 @@ public static float evaluatePolynomial(float[] coeff, float[] data, boolean over
  * @param  weights the curve you want to smooth
  * @return true if weights is modified by fitted curve.
```
deerzq: update the comment about return value

jq: Thanks for review, updated!

core/src/main/java/com/airbnb/aerosolve/core/function/Function.java
```diff
@@ -26,7 +26,11 @@
 
   void resample(int newBins);
 
-  void smooth(double tolerance);
+  // for function not support smooth, just return 0
+  // for function support smooth, return errAndCoeff
+  double smooth(double tolerance);
+  // for function support smooth, return errAndCoeff percentage
+  double smoothByTolerancePercentage(double tolerancePercentage);
```
deerzq: instead of having two different smoothing functions, it might be better to make the percentage an input parameter in `smooth`, since this is in function interface, all the functions would have to implement this even when it's not used

jq: double smooth(double tolerance, boolean isPercentage); ?

jq:
@deerzq thanks for review, issue addressed

deerzq:
LGTM

PR #193, opened 2016-05-28
Fix NDTreePipeline

jq:
1. make sure no dense function won't crash
2. use order of element in the point array as the axis order, oder by largest corresponding delta is copy from KDTree, only make sense if features are same, if features are different, it could result only cut in feature with bigger width. 
3. able to set different size for different dimension
4. add minLeafWidthPercentage, if leaf width is < this number, make it a leaf. this is to reduce number of leaves for highly uneven data.
5. update children's min/max according to parent's split value, this is to share more nodes in Multi-dimension spline.

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
@@ -34,6 +34,9 @@ public class AdditiveModel extends AbstractModel implements Cloneable {
   private Map<String, Function> getOrCreateDenseWeights() {
     if (denseWeights == null) {
       denseWeights = weights.get(DENSE_FAMILY);
+      if (denseWeights == null) {
+        denseWeights = weights.put(DENSE_FAMILY, new HashMap<>());
+      }
     }
     return denseWeights;
   }
@@ -48,7 +51,6 @@ public class AdditiveModel extends AbstractModel implements Cloneable {
     float sum = 0;
     if (denseFeatures != null && !denseFeatures.isEmpty()) {
       Map<String, Function> denseWeights = getOrCreateDenseWeights();
-      if (denseWeights == null) return 0;
       for (Map.Entry<String, List<Double>> feature : denseFeatures.entrySet()) {
         String featureName = feature.getKey();
         Function fun = denseWeights.get(featureName);
@@ -276,15 +278,13 @@ public class AdditiveModel extends AbstractModel implements Cloneable {
     // update with lInfinite cap
     if (denseFeatures != null && !denseFeatures.isEmpty()) {
       Map<String, Function> denseWeights = getOrCreateDenseWeights();
-      if (denseWeights != null) {
-        for (Map.Entry<String, List<Double>> feature : denseFeatures.entrySet()) {
-          String featureName = feature.getKey();
-          Function func = denseWeights.get(featureName);
-          if (func == null) continue;
-          float[] val = FunctionUtil.toFloat(feature.getValue());
-          func.update(-gradWithLearningRate, val);
-          func.LInfinityCap(cap);
-        }
+      for (Map.Entry<String, List<Double>> feature : denseFeatures.entrySet()) {
+        String featureName = feature.getKey();
+        Function func = denseWeights.get(featureName);
+        if (func == null) continue;
+        float[] val = FunctionUtil.toFloat(feature.getValue());
+        func.update(-gradWithLearningRate, val);
+        func.LInfinityCap(cap);
       }
     }
   }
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
```diff
@@ -48,9 +48,11 @@ public float scoreDenseFeatures(Map<String, List<Double>> denseFeatures) {
     float sum = 0;
     if (denseFeatures != null && !denseFeatures.isEmpty()) {
       Map<String, Function> denseWeights = getOrCreateDenseWeights();
+      if (denseWeights == null) return 0;
```
saurfang: I thought the `getOrCreateDenseWeights` guarantee this won't be null?

training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
```diff
@@ -61,16 +64,33 @@ object NDTree {
...
+                  deltas: Array[Double],
+                  node: NDTreeNode,
+                  minLeafWidthPercentage: Double):Int = {
+    for(i <- 1 to deltas.length) {
+      val axis = (lastAxis + i) % deltas.length
+      val rootWidth = node.getMax.get(axis) - node.getMin.get(axis)
+      val width = deltas(axis)
+      if (rootWidth > 0 && width > 0 &&
+          minLeafWidthPercentage * rootWidth <= width) {
+        return axis
```
saurfang: IIRC `return` in Scala is more expensive especially in for loop, because Scala translate for loop into `flatMap`. You can try `find` function on a collection instead which returns the first occurrence the statement is true.

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
```diff
@@ -141,7 +148,19 @@ object NDTreePipeline {
         val result: ((String, String), Either[NDTreeModel, FeatureStats]) = a match {
           case Left(y) => {
             // build tree
-            val model = NDTree(paramsBC.value.options, y.toArray).model
+            val dimension = y(0).length
+            val minLeafWidthPercentage: Double = if (dimension == 1) {
+              1/(2 ^ paramsBC.value.maxTreeDepth1Dimension)
```
saurfang: this will likely return integer rather than double

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/NDTreePipeline.scala
```diff
@@ -141,7 +148,19 @@ object NDTreePipeline {
         val result: ((String, String), Either[NDTreeModel, FeatureStats]) = a match {
           case Left(y) => {
             // build tree
-            val model = NDTree(paramsBC.value.options, y.toArray).model
+            val dimension = y(0).length
```
saurfang: there is no significance of the first element, right? You are just peeking in a random node to find out the dimension?

training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
```diff
@@ -100,36 +120,39 @@ object NDTree {
...
-          getSplitUsingMedian(points, indices, axisIndex)
-        case SplitType.SurfaceArea =>
-          getSplitUsingSurfaceArea(points, indices, bounds, axisIndex)
-      }
-
-      if (split.leftIndices.length <= options.minLeafCount &&
-          split.rightIndices.length <= options.minLeafCount) {
-        log.debug(s"${split.splitValue} leftIndices ${split.leftIndices.length} ${split.rightIndices.length}")
+      // TODO Choose axisIndex with largest corresponding delta as option: deltas.zipWithIndex.maxBy(_._1)._2
+      val axisIndex:Int = getNextAxis(lastAxis, deltas, nodes(0), options.minLeafWidthPercentage)
```
saurfang: Did you mean to pop the head in the process? as is, would `nodes(0)` always point to the same node in subsequent calls?

saurfang:
It would be great to have more documentation on high level description of how trees are built, what assumptions are made, and what control users have over the way it is being built. Right now, it seems the implementation has become powerful enough that these things start to spread out and I find it hard to keep a big picture.

training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
jq: @saurfang Thanks for review! updated and comments added. For this for loop case, if I use find,     List.range(1, deltas.length).find()
the logic in getNextAxis seems hard to bind in, since find just return the i, then I need to construct axis from i, but inside find, I need to construct it too, am I missing anything?

and Yes, I agree, it needs a  high level description, I will do it right after model finish, I am still research what's the right way to construct the tree.... in a debug mode.

saurfang: Roughly I think you can do
```scala
      val deltas = (1 to 10).toArray
      val lastAxis: Int = 8
      deltas.indices.view.map {
        i =>
          println(s"axis for $i")
          (lastAxis + i) % deltas.length
      }.find {
        axis =>
          println(s"looking at $axis")
          val width = deltas(axis)
          width < 3
      }
```
Notice the `view` function after `indices` converts the collection to be lazy and only does map/find one element at a time as oppose to assemble the entire axis array (if this is something you are concerned about.) Otherwise I recommend you leave it out to avoid confusion.

saurfang:
LGTM

deerzq:
LGTM

PR #194, opened 2016-05-31
Infer version from git tag/describe

saurfang:
This changes the published version of Aersolve to be inferred via Git tag. The primary motivation is that it will attached Git commit (and optionally `.dirty` if working copy is dirty) to the end of version string and mark the status as integration. This would allow us to published locally and test the implementation without publishing artifacts to Bintray.

The side effect is that we will need to look directly at Git tag when tagging new version (versus the version is documented in `gradle.properties`. You can also do:
```
$ gradlew :core:printVersion
:core:printVersion
0.1.67-3-g3e1e48e
```
which would indicate last Git tag is `0.1.67`. If this is acceptable, we can merge it and I will update documentation on how to publish and test Aerosolve locally.

@jq @deerzq @spencerde @yolken

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,9 @@ out/
 # NetBeans
 .nb-gradle/
 .solano*
+
+gradle/
+metastore_db/
+derby.log
+demo/income_prediction/output
+demo/income_prediction/src/main/resources/adult.*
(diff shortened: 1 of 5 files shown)


jq:
👍

deerzq:
LGTM

PR #196, opened 2016-06-01
scala pow should use math.pow, and maxTreeDepth1Dimension + 1 so that…

jq:
… maxTreeDepth1Dimension has the same depth meaning as maxTreeDepthPerDimension

@deerzq @saurfang


deerzq:
LGTM

PR #197, opened 2016-06-01
retire maxMinLeafNodesDivByStd due to std is not a good signal to rep…

jq:
…lace dynamic bucket with original spline. also return Array[NDTreeNode] instead of model

@deerzq


deerzq:
if this is not working well, shall retire this entirely? I mean clean up related code in: https://github.com/airbnb/aerosolve/blob/master/core/src/main/java/com/airbnb/aerosolve/core/models/NDTreeModel.java

jq:
Thanks for review! updated

deerzq:
LGTM

PR #198, opened 2016-06-01
[generic-pipeline] add training fraction parameter

deerzq:
to: @jq @saurfang 
cc: @asaluja

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
@@ -80,16 +80,9 @@ object GenericPipeline {
     val inputPattern = cfg.getString("input")
     val subsample = cfg.getDouble("subsample")
     val modelConfig = cfg.getString("model_config")
+    val isTrainingFunc = getIsTrainingFunc(config, isTrainingOpt)
 
-    val input = if (isTrainingOpt.isDefined) {
-      getExamples(sc, inputPattern)
-        .filter(isTrainingOpt.get)
-    } else {
-      val trainingFrac = Try(cfg.getDouble("training_fraction"))
-        .getOrElse(DEFAULT_TRAINING_FRACTION)
-      getExamples(sc, inputPattern)
-        .filter(isTraining(trainingFrac))
-    }
+    val input = getExamples(sc, inputPattern).filter(isTrainingFunc)
 
     val filteredInput = input.sample(false, subsample)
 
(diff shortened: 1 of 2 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
```diff
@@ -509,14 +529,14 @@ object GenericPipeline {
       .map(x => hiveTrainingToExample(x, schema, isMulticlass))
   }
 
-  def isTraining(examples : Example) : Boolean = {
+  def isTraining(frac: Double = DEFAULT_TRAINING_FRACTION)(examples : Example) : Boolean = {
     // Take the hash code mod 255 and keep the first 16 as holdout.
-    (examples.toString.hashCode & 0xFF) > 16
+    (examples.toString.hashCode & 0xFF) > 255 - Math.floor(255 * frac)
   }
```
saurfang: Interesting idea! Is there a reason why we don't use just `examples.hashCode` directly?

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
```diff
@@ -30,6 +30,7 @@ import scala.util.{Random, Try}
 object GenericPipeline {
   val log: Logger = LoggerFactory.getLogger("GenericPipeline")
   val LABEL = "LABEL"
+  val DEFAULT_TRAINING_FRACTION = 0.9373
```
saurfang: Maybe add a comment this magic number is `15/16`?

asaluja:
@deerzq thanks for doing this!

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
deerzq: en, not sure why hector did this

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
deerzq: updated!

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
asaluja: not sure, but isn't this just a performance hack?

jq:
👍

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
saurfang: What `toString` on the Example class does, if I am not wrong, is producing something like `Example@[hex hashcode]`. So `examples.toString.hashCode` is like performing hashing twice on the object, which is probably tiny performance loss. In terms of correctness, this might not be too big of an issue but I would expect Thrift object hashcode to be more uniformly distributed ever so slightly.

PR #199, opened 2016-06-02
[Evaluation] update binary evaluation output

Patrick1860:
After chatting with some Aerosolve users including Cassie, we found the evaluation printout of the binary classification can be improved by.
1) the original TRAIN_PREC@RECALL is supposed to be used to select a threshold fit the specific needs of a user, however, the threshold itself is not printed. So this PR adds threshold in the printout.
2) the wording of TRAIN_PREC@RECALL is not clear. In the text part, precision is listed first. But in the value part, recall is printed first. So this PR also addressed it.
3) change HOLD_PR_AUC to more explicit PRECISION_RECALL_AUC as it could confuse users

@deerzq @jq @sanglifeng


deerzq:
LGTM

PR #200, opened 2016-06-05
Support CSV data sources in GenericPipeline

yolken:
to: @deerzq 

This change allows for CSV data sources in the ```GenericPipeline```. In particular, if the pipeline encounters a query with a reference to ```com.databricks.spark.csv```, it will use a local ```SQLContext``` instead of a ```HiveContext```. Queries run by the former can use the databricks spark-csv library to load a temporary table from a CSV and then execute a query on the loaded data. The table schema is inferred from the header row and values in each column.

Here's an example:

```
generic_hive_query : """
  create temporary table credit_defaults
  using com.databricks.spark.csv
  options (
    path          "file:///home/my_username/my_data_file.csv",
    header        "true",
    inferSchema   "true"
  );
  select
    limit_bal     as i_limitbal,
    sex           as i_sex,
    education     as i_education,
    marraige      as i_marraige,
    label         as LABEL
  from credit_defaults
"""
```

### DIFF
--- a/training/build.gradle
+++ b/training/build.gradle
@@ -35,6 +35,7 @@ dependencies {
   compile 'com.fasterxml.jackson.module:jackson-module-scala_2.10:2.4.2'
   compile 'joda-time:joda-time:2.5'
   compile 'org.apache.hadoop:hadoop-client:2.2.0'
+  compile 'com.databricks:spark-csv_2.10:1.4.0'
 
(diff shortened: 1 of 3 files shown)


deerzq:
Thanks for adding this! probably adding the config example in a config, then LGTM

yolken:
@deerzq Done. Thanks.

PR #201, opened 2016-06-05
Improve NDTree Performance and  obey min leaf count rules

jq:
1. FloatToStringTransform so that you can turn specified float feature to string feature
2. use quick select for median
3. make sure NDTree obey the min leaf count for both children

@deerzq @saurfang @christhetree


deerzq:
LGTM

christhetree:
Awesome adding the quick select! :)

PR #203, opened 2016-06-06
Add dump_forest section to generic pipeline example config

yolken:
to: @deerzq 

Content:
- Add dump_forest section to generic pipeline example config; this is required to dump out a dot file for forest models


deerzq:
LGTM

PR #204, opened 2016-06-07
Make forest model more like Randomforest

Patrick1860:
adding a max_features parameter so that when evaluating a split in a decision tree, it will only consider a random subset of the features. It is a key feature to make Randomforest different from a tree bagging model currently available in Aerosolve. 

@deerzq @jq

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
@@ -72,6 +72,15 @@ object DecisionTreeTrainer {
         .filter(x => x.contains(rankKey))
         .take(candidateSize)
 
+    val ex = examples(0)
+    val numFeatures = LinearRankerUtils.getNumFeatures(ex, rankKey)
+    val maxFeatures : Int = config.getString(key + ".max_features") match {
+      case "all" => numFeatures
+      case "sqrt" => math.sqrt(numFeatures).ceil.toInt
+      case "log2" => math.max(1, (math.log(numFeatures) / math.log(2)).ceil.toInt)
+      case _ => numFeatures
+    }
+
     val stumps = new util.ArrayList[ModelRecord]()
     stumps.append(new ModelRecord)
 
@@ -83,6 +92,7 @@ object DecisionTreeTrainer {
       maxDepth,
       rankKey,
       rankThreshold,
+      maxFeatures,
       numTries,
       minLeafCount,
       SplitCriteria.splitCriteriaFromName(splitCriteriaName)
@@ -102,6 +112,7 @@ object DecisionTreeTrainer {
       maxDepth : Int,
       rankKey : String,
       rankThreshold : Double,
+      maxFeatures : Int,
       numTries : Int,
       minLeafCount : Int,
       splitCriteria : SplitCriteria.Value) : Unit = {
@@ -114,6 +125,7 @@ object DecisionTreeTrainer {
       examples,
       rankKey,
       rankThreshold,
+      maxFeatures,
       numTries,
       minLeafCount,
       splitCriteria
@@ -144,6 +156,7 @@ object DecisionTreeTrainer {
       maxDepth,
       rankKey,
       rankThreshold,
+      maxFeatures,
       numTries,
       minLeafCount,
       splitCriteria
@@ -157,6 +170,7 @@ object DecisionTreeTrainer {
       maxDepth,
       rankKey,
       rankThreshold,
+      maxFeatures,
       numTries,
       minLeafCount,
       splitCriteria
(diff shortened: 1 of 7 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
```diff
@@ -72,6 +72,15 @@ object DecisionTreeTrainer {
         .filter(x => x.contains(rankKey))
         .take(candidateSize)
 
+    val ex = examples(0)
+    val numFeatures = LinearRankerUtils.getNumFeatures(ex, rankKey)
```
deerzq: For the pricing model, different examples always have different number of features which is a small subset of the total number of features we have, for example `San Francisco` might be a missing feature for a listing in `Tokyo`. Due to this reason, I realized there is already randomness in the way we are choosing feature split candidates here. But I guess this change is useful for cases where all examples have a full set of features.

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
```diff
@@ -72,6 +72,15 @@ object DecisionTreeTrainer {
...
+    val maxFeatures : Int = config.getString(key + ".max_features") match {
```
deerzq: It's better to use `Try().getOrElse` to get the training parameter, otherwise this will break some of our pricing jobs that are using decision tree model

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
```diff
@@ -72,6 +72,15 @@ object DecisionTreeTrainer {
...
+      case "all" => numFeatures
+      case "sqrt" => math.sqrt(numFeatures).ceil.toInt
+      case "log2" => math.max(1, (math.log(numFeatures) / math.log(2)).ceil.toInt)
+      case _ => numFeatures
```
deerzq: by default, we should just don't do sampling, you might just use `Int.MaxValue` here, since different examples may have different `numFeatures`

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
Patrick1860: for missing value, the generic pipeline has a missing family, will the current implementation of getNumFeatures function still count the number of element in the missing family and get a full number of the feature or `numFeatures ` will depend on the sample ex sampled?

yay, the model should function as data dependent

Patrick1860:
@deerzq thanks for the comments. Addressed and PTAL

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
deerzq: actually it's not really 'missing'. for example, all examples may have a feature called 'dim_market' in the hive row. but inherently, if there are 1000 markets, this feature will become 1000 different features, then for listing in 'San Francisco', the feature 'Tokyo' is always missing and won't be put under 'MISS' family since 'dim_market' feature is not missing.

deerzq:
LGTM

training/src/main/scala/com/airbnb/aerosolve/training/ForestTrainer.scala
```diff
@@ -43,6 +43,15 @@ object ForestTrainer {
         .map(x => Util.flattenFeature(x.example(0)))
         .filter(x => x.contains(rankKey))
         .coalesce(numTrees, true)
+
+    val ex = examples.take(1)(0)
+    val numFeatures = LinearRankerUtils.getNumFeatures(ex, rankKey)
+    val maxFeatures : Int = Try(config.getString(key + ".max_features")).getOrElse("all") match {
+      case "all" => Int.MaxValue
+      case "sqrt" => math.sqrt(numFeatures).ceil.toInt
+      case "log2" => math.max(1, (math.log(numFeatures) / math.log(2)).ceil.toInt)
```
jq: this seems dup with DecisionTreeTrainer

training/src/main/scala/com/airbnb/aerosolve/training/DecisionTreeTrainer.scala
Patrick1860: makes sense. Thanks for the explanation.

training/src/main/scala/com/airbnb/aerosolve/training/ForestTrainer.scala
Patrick1860: because the `ForestTrainer` currently directly calls the buildTree function in decisionTree training. I think it is better to have `buildTree` function just take a int argument so it works independent of what the config value could be for different models. The `ForestTrainer` and DecisionTreeTrainer currently have the same logic in term of parsing the max_features config but they can be different later on. Let me know if makes sense to you. @jq

Patrick1860: If there is no more comments I will merge it tonight.

PR #207, opened 2016-06-09
Decay Learning Rate

jq:
learning rate decay each iteration 
learning_rate = max(learning_rate_decay*learning_rate, min_learning_rate)

@deerzq @saurfang


deerzq:
LGTM

PR #209, opened 2016-06-10
params is small, let's not use it directly without broadcast to reduc…

jq:
…e the error such as Failed to get broadcast_0_piece0 of broadcast_0

@deerzq @saurfang


deerzq:
LGTM

PR #210, opened 2016-06-10
Allow string cross to specify set of cross features.

jq:
@deerzq @saurfang

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatFamilyToDenseTransform.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatFamilyToDenseTransform.java
@@ -4,11 +4,16 @@ import com.airbnb.aerosolve.core.FeatureVector;
 import com.airbnb.aerosolve.core.util.Util;
 import com.typesafe.config.Config;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 
 /**
  * cross one float family with the other and out a 2D dense feature.
+ * if fieldsName2 missing, this is a self cross
+ * self cross use key's alphabetical order to determine order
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/airbnb/aerosolve/core/util/Util.java
```diff
@@ -379,6 +381,22 @@ public int compare(DebugScoreDiffRecord e1, DebugScoreDiffRecord e2) {
     return debugDiffRecord;
   }
 
+  public static <T> Set<T> getIntersection(Set<T> a, Set<T> b) {
+    if (a == null) {
+      return b;
```
deerzq: This logic is not consistent with the name `getIntersection`, since by definition, this should return null if one of the set is empty. We may let this `getIntersection` do what it should do and move this logic to the cross transform.

jq:
@deerzq thanks for review, issue addressed.

core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatFamilyToDenseTransform.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * cross one float family with the other and out a 2D dense feature.
+ * if fieldsName2 missing, this is a self cross
+ * self cross use key's alphabetical order to determine order
+ * so you can add 1_ 2_ in front of features to manipulate order.
+ */
+public class FloatFamilyToDenseTransform implements Transform {
```
deerzq: rename this? it does not reflect the fact that you are doing some cross internally

deerzq: you may also want the name to reflect the fact that the output is a 2D dense feature.

PR #212, opened 2016-06-12
use System.currentTimeMillis to make sample truly random

jq:
@deerzq @saurfang


deerzq:
LGTM

saurfang:
The default seed is already random by default, no? https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L476

jq:
@saurfang  you are right, it is random, val random = new Random() and Random () is inited with random seed.

PR #213, opened 2016-06-12
replace min_leaf_count with min_leaf_count_percent and add min_leaf_v…

jq:
…alue_percent so that we adjust leaf size relative to sample size.
also no longer use broadcast since param is very small.

@deerzq @saurfang

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
@@ -25,11 +25,11 @@ object NDTree {
   val defaultMultiDimensionalSplitType = SplitType.Median
 
   case class NDTreeBuildOptions(
-      maxTreeDepth: Int,
-      minLeafCount: Int,
-      // (leaf's max - min)/(root's max - min)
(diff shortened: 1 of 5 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/NDTree.scala
```diff
@@ -25,11 +25,11 @@ object NDTree {
...
   case class NDTreeBuildOptions(
-      maxTreeDepth: Int,
-      minLeafCount: Int,
-      // (leaf's max - min)/(root's max - min)
-      minLeafWidthPercentage: Double = 0,
-      splitType: SplitType.Value = SplitType.Unspecified)
+    maxTreeDepth: Int,
+    minLeafCount: Int,
+    // (leaf's max - min)/(root's max - min)
+    minLeafValuePercent: Double = 0,
```
deerzq: what's the difference between `minLeafWidthPercentage`?

jq: minLeafWidthPercentage is minLeafValuePercent, change the name

deerzq:
LGTM

PR #217, opened 2016-06-14
Revert "convert family and feature names to lower case"

jq:
This reverts commit 866f14539fb27267351a5cbfbb391366e33ea3e4.

still need this so that lower case label can be accepted as label, and it wont' break existing model or scoring since it is label, and it accept compareToIgnoreCase of LABEL

@deerzq


deerzq:
LGTM

PR #218, opened 2016-06-14
add back minLeafCount, since some feature may have very few samples, …

jq:
…just control by relative value is not enough.

@deerzq @saurfang


deerzq:
LGTM

PR #221, opened 2016-06-23
set feature families using spline instead of dynamic bucketing

jq:
@deerzq @saurfang


deerzq:
LGTM

saurfang:
tests fail to compile. LGTM otherwise

jq:
Thanks for review, Updated Unit test

PR #223, opened 2016-06-24
Fix Point function mistakenly init to weight 1

jq:
@deerzq @saurfang


deerzq:
LGTM

spritehc:
Change title to weight `0`?

PR #225, opened 2016-07-08
allow dump model to update given hive table partition

deerzq:
to: @jq @yolken


jq:
👍   please also update the generic pipeline hackpad

deerzq:
sure, will update the document

PR #227, opened 2016-07-13
Add eval_test option

spritehc:
`eval_test` was an option in income pipeline but not part of Genericpipeline. The goal is to be able to evaluate the performance of model on a separate test data other than `hold_out` data as specified in `eval_test`.

Example usage:

```
make_testing {
  hive_query : ${generic_hive_query}" from "${training_table}${testing_filter}
  num_shards : 20
  output : ${testing_data}
}

eval_test {
  input : ${testing_data}
  subsample : ${eval_subsample}
  bins : 100
  model_config : ${model_config}
  is_probability : false
  is_regression : false
  metric_to_maximize : "!HOLD_F1"
  model_name : ${model_name}
}
```

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
@@ -33,16 +33,15 @@ object GenericPipeline {
   val LABEL = "LABEL"
   val DEFAULT_TRAINING_FRACTION = 0.9373 // 0.9373 = (255 - 16) / 255
 
-  def makeTrainingRun(sc: SparkContext, config: Config) = {
-    val cfg = config.getConfig("make_training")
+  def makeExampleRun(sc: SparkContext, cfg: Config) = {
     val query = cfg.getString("hive_query")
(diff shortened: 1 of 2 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
```diff
@@ -48,6 +48,20 @@ object GenericPipeline {
       .saveAsTextFile(output, classOf[GzipCodec])
   }
 
+  def makeTestingRun(sc: SparkContext, config: Config) = {
```
deerzq: This is actually doing the same thing as `makeTrainingRun` it's just the config is different. I feel we can have some thing like `GenericPipeline.makeExampleRun(sc, config)`, then we can share some code.

jq:
please update the config too https://github.com/airbnb/aerosolve/blob/master/training/src/main/resources/demo_train.conf

jq:
you have compile error...

spritehc:
@jq Sorry, I wasn't actively working on the PR. updated

jq:
@spritehc nice, can you update the sample config too?

PR #228, opened 2016-07-13
Adds periods where appropriate

ClimbsRocks:
It was weird to have these two sentences end without periods: 
image
image

And this period was included in the hyperlink text, so I moved it to come after the hyperlink:
image


deerzq:
LGTM

PR #229, opened 2016-07-18
Optimize AdditiveModelTrainer using sparse representation

saurfang:
This is a huge one. Thanks for reviewing it in advance. You might want to examine commit by commit and also add `&w=1` or `?w=1` to the URL to ignore white spaces in the diff due to reformatting.

Aside a few minor optimizations mentioned in the commits, the major changes are:
1. Add `SparseLabeledPoint` for space efficient representation of feature vector
2. Add validation split into model trainer for better understanding of model convergence

## Sparse Representation
It shall be noted this implementation is not ideal in any sense but as a compromise to best fit in Aerosolve existing architecture without rewriting all the transforms.

`SparseLabeledPoint` store feature vector in the form of
```java
    public final int[] indices;
    public final float[] values;
    public final int[] denseIndices;
    public final float[][] denseValues;
```
where float and string features are stored in `indices,values` and dense features are stored in `denseIndices,denseValues`. The index of each feature is backed by `featureIndexer` in AdditiveModel of type `Map<String, Map<String, Integer>>`.

So instead of
```
Map(
  'family1' -> Map(
    `feature1' -> value1,
    `feature4' -> value2,
   ...),
  ...
)
```
we store features as
```
indices: [1, 4, ...]
values: [1, 2, ...]
```
Furthermore, to support efficient model updates with sparsely represented feature vector, we collect all features into `Function[] weightVector` where features are directly indexed by their corresponding index in the `featureIndexer`.

## Validation Split
Because SGD loss is rather jittery in short window frame and we perform averaging of model weights among other regularizations during each iterations, training loss appears to be a poor proxy to model convergence.
This PR adds an optional `isTraining` function to split training data into training and validation. After each iteration, validation loss is calculated without updating the model itself.
This gives a much clearer guidance to both learning rate specification, its appropriate decay rate, early stopping, and if model of any iterations might have been overfitted.
The final model is picked from the iteration with lowest validation loss instead.

## Speed Optimization
With validation loss, we were able to experiment with parameters and sampling technique with much better confidence. 
This PR also adds option to `sample_by_partition` which take a sample of partitions. This played nicely when there are lots of partitions and data is randomly distributed when they are pulled. It appears model can converge much faster when a smaller sample is chosen and run with a few more iterations.

@deerzq @jq @timyitong

p.s. I will update hackpad once this is merged to outline a few techniques and findings.

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/features/SparseLabeledPoint.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/features/SparseLabeledPoint.java
@@ -1,0 +1,19 @@
+package com.airbnb.aerosolve.core.features;
+
+import lombok.Data;
+
+import java.io.Serializable;
+
+/**
+ * This is a compressed representation of a feature vector indexed via a feature index map. It captures values in
+ * flat representation for space and runtime efficiency during model training.
+ */
+@Data
+public class SparseLabeledPoint implements Serializable {
+    public final boolean isTraining;
+    public final double label;
+    public final int[] indices;
+    public final float[] values;
+    public final int[] denseIndices;
+    public final float[][] denseValues;
+}
--- a/core/src/main/java/com/airbnb/aerosolve/core/features/SparseLabeledPoint.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/features/SparseLabeledPoint.java
@@ -5,15 +5,15 @@ import lombok.Data;
 import java.io.Serializable;
 
 /**
- * This is a compressed representation of a feature vector indexed via a feature index map. It captures values in
- * flat representation for space and runtime efficiency during model training.
+ * This is a compressed representation of a feature vector indexed via a feature index map. It
+ * captures values in flat representation for space and runtime efficiency during model training.
  */
 @Data
 public class SparseLabeledPoint implements Serializable {
-    public final boolean isTraining;
-    public final double label;
-    public final int[] indices;
-    public final float[] values;
-    public final int[] denseIndices;
-    public final float[][] denseValues;
+  public final boolean isTraining;
+  public final double label;
+  public final int[] indices;
+  public final float[] values;
+  public final int[] denseIndices;
+  public final float[][] denseValues;
 }
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
@@ -7,8 +7,8 @@ import com.airbnb.aerosolve.core.ModelRecord;
 import com.airbnb.aerosolve.core.features.SparseLabeledPoint;
 import com.airbnb.aerosolve.core.function.AbstractFunction;
 import com.airbnb.aerosolve.core.function.Function;
-import com.airbnb.aerosolve.core.function.FunctionUtil;
 import com.airbnb.aerosolve.core.util.Util;
+
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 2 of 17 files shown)


timyitong:
@saurfang: thanks for making this exciting change! Should simplify our internal logics a lot. A few higher level ideas before going into details:

Do you think it's worth considering further wrap the current ```SparseLabeledPoint``` interface into something similar as https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/mllib/regression/LabeledPoint.scala#L36

The difference is that Spark ML is leveraging [Breeze](https://github.com/scalanlp/breeze) for vector representation. It can handle nicely with both sparse/dense vector, a bunch of linear algebra computations and even a few convex optimizers. Though, I do believe this doesn't have to happen in this PR as it is already making a significant change complicated enough.

timyitong:
Another thing is about denseIndices and denseValues; now it looks like more of a special structure tweak introduced due to the format ```Example``` is assuming. Should dense vector also be just stored into the sparse array? (Assuming the majority of features is sparse, this might be okay).

saurfang:
@timyitong I originally planned to do that but then I found I need to make a few modifications to make it work (i.e. training split, dense features and etc). Since there isn't a place where we can leverage Breeze for now, I left it out for this PR. I agree we should migrate towards that.

I think dense features are supposedly reserved for multi-dimensional spline, which I think is still experimental. Right now this makes it easier to drop if we don't find it useful but yes I think we could eventually migrate to an unified representation.

core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
```diff
@@ -16,305 +17,376 @@
...
-  // whole dense features belongs to feature family DENSE_FAMILY
-  private Map<String, Function> denseWeights;
-
-  private Map<String, Function> getOrCreateDenseWeights() {
-    if (denseWeights == null) {
-      denseWeights = weights.get(DENSE_FAMILY);
-      if (denseWeights == null) {
-        denseWeights = weights.put(DENSE_FAMILY, new HashMap<>());
-      }
+    public static final String DENSE_FAMILY = "dense";
```
jq: switch to 4 space? @deerzq

training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerUtils.scala
```diff
@@ -119,74 +123,198 @@ object LinearRankerUtils {
...
+    * Each observation is further flattened to [[SparseLabeledPoint]] format where features are indexed
+    * by a specified [[AdditiveModel]].
+    */
+  def makePointwiseFloatVector(examples: RDD[Example],
+                               transformer: Transformer,
+                               params: AdditiveModelTrainer.AdditiveTrainerParams,
+                               model: AdditiveModel,
+                               isTraining: Example => Boolean = _ => true,
+                               groupSize: Int = 100): RDD[SparseLabeledPoint] = {
+    val assemblerTimer = examples.sparkContext.accumulator(0L, "pointAssembler")
```
jq: this is for debug speed?

core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
deerzq: perhaps keep this consistent with our internal format would be better

jq:
so it move lookups from training, to example => SparseLabeledPoint, performance should be same?

training/src/main/scala/com/airbnb/aerosolve/training/LinearRankerUtils.scala
saurfang: Correct and the time spent here is actually ineligible and would be really great if we rewrite the transform altogether in the future.

core/src/main/java/com/airbnb/aerosolve/core/models/AdditiveModel.java
saurfang: Sticking with internal format sounds good to me. I'll update the PR.

saurfang:
Lookup itself can be slightly faster due to elimination of double hashmap lookup, hash collision, string comparison, and nested map traversal. The shuffle size itself is 2-3 times smaller which saves a lot of time (but like we said above the assembling itself is not free) and since this is more memory efficient it also saves a few GC pauses. All in all the runtime speedup is okay (about ~25% faster).

More interestingly though is the `sample_by_partition` option, which makes the model converges much faster in wall clock time (about 50% faster) i.e. train more iterations but with much less fraction of data during each iterations. The validation loss gives a good measuring stick of convergence in this case.

training/src/test/scala/com/airbnb/aerosolve/training/TrainingTestHelper.scala
```diff
@@ -180,7 +180,7 @@ object TrainingTestHelper {
     example
   }
 
-  def makeClassificationExamples = {
+  lazy val makeClassificationExamples = {
```
jq: This is nice!

jq:
search sample_by_partition found nothing, @saurfang

saurfang:
@jq sorry. It's called `partition_subsample` or `paritionSubsample`. This is a coarse level sampling in addition to the partition wise sampling we do via `subsample`.

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -154,41 +380,56 @@ object AdditiveModelTrainer {
...
     val params = sgdParams.params
-    val data = input(params.subsample)
-      .coalesce(params.numBags, params.shuffle)
+    val data = if (lossOnly) {
+      input(1.0) // input has already been sampled during split
+    } else if (params.partitionSubsample < 1) {
+      val numPartitions = input(1.0).partitions.length
+
+      val sampledPartitions = random.shuffle((0 until numPartitions).toList)
+        .take((numPartitions * params.subsample).toInt)
```
jq: should it be partitionSubsample?

saurfang: good catch :+1: thanks!

jq:
👍  @saurfang  suggest add the new config to demo app, may catch bugs

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
```diff
@@ -213,80 +214,74 @@ object TrainingUtils {
...
+      .agg(
+        count(lit(1)) as "count",
+        min($"value"),
+        max($"value"),
+        sum($"value"),
+        // TODO: use built-in stdev function in Spark 1.6
+        sum($"value" * $"value")
+      )
+      // ignore features present in less than minCount examples
+      .where($"count" >= minCount)
```
deerzq: so moving this to sparkSQL will make this faster?

core/src/main/java/com/airbnb/aerosolve/core/transforms/ListTransform.java
```diff
@@ -34,4 +35,13 @@ public void doTransform(FeatureVector featureVector) {
       transform.doTransform(featureVector);
     }
   }
+
+  @Override
+  public void doTransform(Stream<FeatureVector> featureVectors) {
```
deerzq: why `Stream` instead of something like `List`, I feel it can be tricky for debugging when using `Stream` as type of input parameter due to its lazy evaluation. perhaps you can convert this to stream inside?

deerzq:
Looking forward to faster training! overall LGTM. As jq mentioned, please make sure to test with more comprehensive examples (e.g. the pricing model) before merging.

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
```diff
@@ -213,80 +214,74 @@ object TrainingUtils {
...
+      .map {
+        case Row(family: String, feature: String, count: Long, min: Double, max: Double, sum: Double, sqSum: Double) =>
+          ((family, feature), FeatureStatistics(count, min, max, sum / count, (sqSum - sum * sum / count) / (count - 1)))
```
timyitong: Will it occur that ```minCount == 0 and count = 1```?

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
```diff
@@ -115,7 +116,7 @@ object TrainingUtils {
       return None
```
timyitong: remove this return also?

core/src/main/java/com/airbnb/aerosolve/core/features/SparseLabeledPoint.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.Data;
+
+import java.io.Serializable;
+
+/**
+ * This is a compressed representation of a feature vector indexed via a feature index map. It
+ * captures values in flat representation for space and runtime efficiency during model training.
+ */
+@Data
+public class SparseLabeledPoint implements Serializable {
```
timyitong: Maybe a safe check during construction that indices/values are having the same length?

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
saurfang: Slightly faster but not as significant as I had hoped. Since code itself is more readable, I decided to kept it.

core/src/main/java/com/airbnb/aerosolve/core/transforms/ListTransform.java
saurfang: Oh yes. I did this originally for performance but then found out that `ListTransform` has to collect all the feature vectors so it can operate on them repeatedly, thus defeating the purpose. I'm going to change it to `Iterable` instead.

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
saurfang: I think mean and/or variance will be not a number and such feature will probably be dropped anyway.

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
saurfang: Good call. All return statement in Scala has been eradicated.

PR #231, opened 2016-07-22
Write metrics to result file during eval

abaek:
If a user specifies a `result` parameter in the eval part of their config, then it will write all metrics to that file.

To: @jq @yinlou

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
@@ -3,6 +3,7 @@ package com.airbnb.aerosolve.training
 import org.slf4j.{Logger, LoggerFactory}
 import org.apache.spark.rdd.RDD
 import com.airbnb.aerosolve.core.EvaluationRecord
+import com.airbnb.aerosolve.training.pipeline.ResultUtil
 import org.apache.spark.SparkContext._
 import scala.collection.{mutable, Map}
 import scala.collection.mutable.{ArrayBuffer, Buffer}
--- a/training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
@@ -20,119 +20,25 @@ object Evaluation {
   def evaluateBinaryClassification(records : RDD[EvaluationRecord],
                                    buckets : Int,
                                    evalMetric : String) : Array[(String, Double)] = {
-    var metrics  = mutable.Buffer[(String, Double)]()
-    var bestF1 = -1.0
-    val thresholds = records.map(x => x.score).histogram(buckets)._1
-    // Search all thresholds for the best F1
-    // At the same time collect the precision and recall.
-    val trainPR = new ArrayBuffer[(Double, Double, Double)]()
-    val holdPR = new ArrayBuffer[(Double, Double, Double)]()
-    trainPR.append((1.0, 0.0, thresholds.max))
-    holdPR.append((1.0, 0.0, thresholds.max))
-
-    for (i <- 0 until thresholds.size - 1) {
-      val threshold = thresholds(i)
-      val tmp = evaluateBinaryClassificationAtThreshold(records, threshold)
-      val tmpMap = tmp.toMap
-      val f1 = tmpMap.getOrElse(evalMetric, 0.0)
-      if (f1 > bestF1) {
-        bestF1 = f1
-        metrics = tmp
-      }
-      trainPR.append((tmpMap.getOrElse("!TRAIN_PRECISION", 0.0), tmpMap.getOrElse("!TRAIN_RECALL", 0.0), threshold))
-      holdPR.append((tmpMap.getOrElse("!HOLD_PRECISION", 0.0), tmpMap.getOrElse("!HOLD_RECALL", 0.0), threshold))
-    }
-
-    metricsAppend(metrics, trainPR, holdPR)
-
-    evaluateBinaryClassificationAUC(records, metrics)
-
-    metrics
-      .sortWith((a, b) => a._1 < b._1)
-      .toArray
+    evaluateBinaryClassificationWithResults(records, buckets, evalMetric, null)
   }
 
   def evaluateBinaryClassification(records : List[EvaluationRecord],
                                    buckets : Int,
                                    evalMetric : String) : Array[(String, Double)] = {
-    var metrics  = mutable.Buffer[(String, Double)]()
-    var bestF1 = -1.0
-    val scores: List[Double] = records.map(x => x.score)
-    val thresholds = getThresholds(scores, buckets)
-    // Search all thresholds for the best F1
-    // At the same time collect the precision and recall.
-    val trainPR = new ArrayBuffer[(Double, Double, Double)]()
-    val holdPR = new ArrayBuffer[(Double, Double, Double)]()
-    trainPR.append((1.0, 0.0, thresholds.max))
-    holdPR.append((1.0, 0.0, thresholds.max))
-
-    for (i <- 0 until thresholds.length - 1) {
-      val threshold = thresholds(i)
-      val tmp = evaluateBinaryClassificationAtThreshold(records, threshold)
-      val tmpMap = tmp.toMap
-      val f1 = tmpMap.getOrElse(evalMetric, 0.0)
-      if (f1 > bestF1) {
-        bestF1 = f1
-        metrics = tmp
-      }
-      trainPR.append((tmpMap.getOrElse("!TRAIN_PRECISION", 0.0), tmpMap.getOrElse("!TRAIN_RECALL", 0.0), threshold))
-      holdPR.append((tmpMap.getOrElse("!HOLD_PRECISION", 0.0), tmpMap.getOrElse("!HOLD_RECALL", 0.0), threshold))
-    }
-
-    metricsAppend(metrics, trainPR, holdPR)
-
-    evaluateBinaryClassificationAUC(records, metrics)
-
-    metrics
-      .sortWith((a, b) => a._1 < b._1)
-      .toArray
+    evaluateBinaryClassificationWithResults(records, buckets, evalMetric, null)
   }
 
-  def evaluateBinaryClassification(records : RDD[EvaluationRecord],
+  def evaluateBinaryClassificationWithResults(records : RDD[EvaluationRecord],
                                    buckets : Int,
                                    evalMetric : String,
                                    resultsOutputPath: String) : Array[(String, Double)] = {
-    var metrics  = mutable.Buffer[(String, Double)]()
-    var bestF1 = -1.0
-    val thresholds = records.map(x => x.score).histogram(buckets)._1
-    var results = mutable.Buffer[(String, Double)]()
-    var holdThresholdPrecisionRecall = mutable.Buffer[(Double, Double, Double)]()
-    var trainThresholdPrecisionRecall = mutable.Buffer[(Double, Double, Double)]()
-
-    // Search all thresholds for the best F1
-    // At the same time collect the precision and recall.
-    val trainPR = new ArrayBuffer[(Double, Double, Double)]()
-    val holdPR = new ArrayBuffer[(Double, Double, Double)]()
-    trainPR.append((1.0, 0.0, thresholds.max))
-    holdPR.append((1.0, 0.0, thresholds.max))
-
-    for (i <- 0 until thresholds.size - 1) {
-      val threshold = thresholds(i)
-      val tmp = evaluateBinaryClassificationAtThreshold(records, threshold, results)
-      val tmpMap = tmp.toMap
-      val f1 = tmpMap.getOrElse(evalMetric, 0.0)
-      if (f1 > bestF1) {
-        bestF1 = f1
-        metrics = tmp
-      }
-      trainPR.append((tmpMap.getOrElse("!TRAIN_PRECISION", 0.0), tmpMap.getOrElse("!TRAIN_RECALL", 0.0), threshold))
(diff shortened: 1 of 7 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
```diff
@@ -139,6 +139,23 @@ object GenericPipeline {
     val metrics = evalCompute(sc, config, cfgKey, isTrainingFunc)
 
     metrics.foreach(x => log.info(if (x._1 contains "THRESHOLD") x._1 else x.toString))
+
+    // Write metrics to result
+    try {
```
jq: please wrap it into class.

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
```diff
@@ -139,6 +139,23 @@ object GenericPipeline {
...
+      val result: String = confi.getConfig(cfgKey).getString("result")
```
jq: you will need the config of output result path, i.e. if it has output_result_path, then save result this way

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
yinlou: agreed. Could be something like this:

def outputResults(records to write): Unit = {
  // log
  records.map(e => log.info(e.toString))
  // Write to hdfs
  Your logic
}

abaek: If I wrap it into a class, im not sure if i should also move the logging there

abaek:
Build doesn't seem to pass, but could you verify that this is the right place to put the code. I can fix the syntax afterwards. @jq

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
```diff
@@ -156,6 +156,7 @@ object GenericPipeline {
     val isRegression = Try(cfg.getBoolean("is_regression")).getOrElse(false)
     val isMulticlass = Try(cfg.getBoolean("is_multiclass")).getOrElse(false)
     val metric = cfg.getString("metric_to_maximize")
+    val resultsOutputPath = Try(cfg.getString("results_output_path"))..getOrElse("")
```
jq: ..

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
```diff
@@ -447,7 +449,8 @@ object GenericPipeline {
       isRegression: Boolean,
       isMulticlass: Boolean,
       metric: String,
-      isTraining: Example => Boolean) : Array[(String, Double)] = {
+      isTraining: Example,
+      resultsOutputPath: String => Boolean) : Array[(String, Double)] = {
```
jq: Example => Boolean and String

training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
```diff
@@ -87,6 +87,24 @@ object Evaluation {
...
+                                   resultOutputPath: String) : Array[(String, Double)] = {
+    val metrics = evaluateBinaryClassification(records, buckets, evalMetric)
+    ResultUtil.writeResults(metrics, resultOutputPath)
+    metrics
+  }
+
+  def evaluateBinaryClassification(records : List[EvaluationRecord],
+                                   buckets : Int,
+                                   evalMetric : String,
+                                   resultOutputPath: String) : Array[(String, Double)] = {
```
jq: this two are same?

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/ResultUtil.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.FileSystem
+
+/*
+ * Writes to result file
+ */
+object ResultUtil {
+  def writeResults(metrics: Array[(String, Double)], resultOutputPath: String) = {
+    val fileSystem = FileSystem.get(new java.net.URI(result), new Configuration())
```
jq: try reuse https://github.com/airbnb/aerosolve/blob/master/training/src/main/scala/com/airbnb/aerosolve/training/pipeline/PipelineUtil.scala#L39

jq:
ok to put it in current places. You may talk to yizheng about the format of result.

training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
abaek: One of them is for List and the other is for RDD

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
abaek: I'm not sure what you mean, I thought this line was correct

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/ResultUtil.scala
abaek: `PipelineUtil.saveAndCommitAsTextFile` didn't have exactly what i needed, but I used `PipelineUtil.writeStringToFile`

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
abaek: nvm, i  understand

abaek:
Failed a test when I ran `gradle test`, but I don't think it was related to my changes

screen shot 2016-07-26 at 3 50 16 pm

abaek:
Actually, working on changing the format, don't review yet please

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
```diff
@@ -156,6 +156,7 @@ object GenericPipeline {
     val isRegression = Try(cfg.getBoolean("is_regression")).getOrElse(false)
     val isMulticlass = Try(cfg.getBoolean("is_multiclass")).getOrElse(false)
     val metric = cfg.getString("metric_to_maximize")
+    val resultsOutputPath = Try(cfg.getString("results_output_path")).getOrElse("")
```
jq: please update the demo with results_output_path, this is a good test for the code too.

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/ResultUtil.scala
```diff
@@ -0,0 +1,19 @@
...
+import com.airbnb.aerosolve.training._
+
+/*
+ * Writes to result file
+ */
+object ResultUtil {
+  def writeResults(metrics: Array[(String, Double)], resultOutputPath: String) = {
+    // Put into a one line json format
+    val elements = metrics.map{ x =>
+      val metricStr = if (x._1 contains "THRESHOLD") x._1 else x.toString
```
jq: move THRESHOLD out, and add Unit test and comments for the output format.

abaek:
I changed the result output format and updated this PR. Apparently you can't have default arguments and overload method names in scala, so I created new methods and resued a lot of code from the old methods.

I still have to work on adding a unit test, and testing this code with the demo.

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
abaek: done

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/ResultUtil.scala
abaek: I added comments for the output format, but I wasn't sure what to test in a unit test

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
jq: seems not yet push? i.e. demo https://github.com/airbnb/aerosolve/tree/master/demo/income_prediction

training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
```diff
@@ -87,6 +88,95 @@ object Evaluation {
       .toArray
   }
 
+  def evaluateBinaryClassification(records : RDD[EvaluationRecord],
+                                   buckets : Int,
+                                   evalMetric : String,
+                                   resultsOutputPath: String) : Array[(String, Double)] = {
+    var metrics  = mutable.Buffer[(String, Double)]()
```
jq: lots of code are dup with the other evaluateBinaryClassification, can you share code between them?

training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
```diff
@@ -197,6 +330,16 @@ object Evaluation {
     metrics.append(("!TRAIN_PRECISION", if (ttp > 0.0) ttp / (ttp + tfp) else 0.0))
     metrics.append(("!TRAIN_F1", 2 * ttp / (2 * ttp + tfn + tfp)))
 
+    if (results != null) {
```
jq: maybe it is better to report it with metrics.append, i.e. metricsAppend(metrics, results, name, value)

training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
abaek: done, code is never repeated

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/GenericPipeline.scala
```diff
@@ -470,7 +473,7 @@ object GenericPipeline {
     } else if (isMulticlass) {
       Evaluation.evaluateMulticlassClassification(records)
     } else {
-      Evaluation.evaluateBinaryClassification(records, bins, metric)
+      Evaluation.evaluateBinaryClassificationWithResults(records, bins, metric, resultsOutputPath)
```
jq: if resultsOutputPath is null, appendMetricAndResult may crash..

abaek: I got rid of `appendMetricAndResult`. Now I am just relying on the metrics buffer and `resultsOutputPath` is only being used in one place here `Evaluation#L78`: 
```
    if (resultsOutputPath != null) {
      ResultUtil.writeResults(resultsOutputPath, metrics.toArray, holdThresholdPrecisionRecall.toArray, trainThresholdPrecisionRecall.toArray)
    }
```

jq: nice, please push the update

abaek: Woops done

training/src/main/scala/com/airbnb/aerosolve/training/pipeline/ResultUtil.scala
```diff
@@ -0,0 +1,113 @@
...
+      0.313343,
+      0.897908,
+      1.190191,
+      1.482474
+    ]
+    }
+ *
+ */
+object ResultUtil {
+  def writeResults(resultsOutputPath: String,
```
jq: suggest you add unit test for ResultUtil in this PR, and make ResultUtil extendable on next PR

jq:
👍

training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
```diff
@@ -20,119 +20,25 @@ object Evaluation {
...
-    thresholdsAppend(trainPR, holdPR, holdThresholdPrecisionRecall, trainThresholdPrecisionRecall)
-    val resultsArray = results.sortWith((a, b) => a._1 < b._1).toArray
-    ResultUtil.writeResults(resultsOutputPath, resultsArray, holdThresholdPrecisionRecall.toArray, trainThresholdPrecisionRecall.toArray)
-
-    metrics
-      .sortWith((a, b) => a._1 < b._1)
-      .toArray
+    // Convert RDD to list
+    val recordsList = List.fromArray(records.collect());
+    evaluateBinaryClassificationWithResults(recordsList, buckets, evalMetric, resultsOutputPath)
```
saurfang: @abaek what was the motivation to call `collect` here? was it just to unify the codepath? I thought there were two evaluations that can compute locally when result is small but distributed via RDD when dataset is large.

abaek: The motivation was to make all of the `evaluateBinaryClassification` share one code path and eliminate duplicate code. When I tried to refactor the code, I had a hard time sharing code between RDD's and List.

@jq mentioned that Evaluation jobs were small enough that collecting the RDD was ok.

saurfang: I see. It makes sense but I'm having trouble collecting them to driver and driver is struggling with the evaluation. I'm going to try making everything follow the RDD codepath instead.

abaek: Oh ok sorry about that! I think I only tested it with Lists.

PR #232, opened 2016-07-27
Allow kdtree bounds update

saurfang:
Adds an `updateBounds` function to `KDTree` so boundary of KDTree node can be updated after a KDTree model is trained. Split is not affected by this update. The boundary information is purely informational.

Reformatted code and applied static code analysis suggestions.

@jq @spencerde


jq:
👍  what static code analysis tool?

saurfang:
I meant the code suggestions IntelliJ gives. Most of them are good advice (alt+enter is the shortcut to take such advice in case you didn't know)

spencerde:
LGTM

PR #233, opened 2016-08-03
add evaluation data source

Patrick1860:
The training and evaluation dataset is usually different and in a lot of the cases need to be separated by time not by random sample. e.g. training is data from 2015 and eval is data from 2016. 

The current Aerosolve config doesn’t support adding evaluation dataset. This PR is to add this option to the user. As suggested by @deerzq , I refactored the makeTraningrun to make it more generic to make both the training and evaluation dataset(optional)

@jq @saurfang

### DIFF
--- a/training/src/main/resources/demo_multiclass_train.conf
+++ b/training/src/main/resources/demo_multiclass_train.conf
@@ -13,10 +13,14 @@ model_version: "a"
 model_type : "full_rank_linear"
 model_config : ${model_type}"_model_config"
 demo_table : "pricing.twenty_news_training_raw"
+# Can use a different table
+demo_table_eval : "pricing.twenty_news_training_raw"
 
 # Where to dump outputs; change prefix based on your HDFS directory layout
 prefix : "hdfs://airfs-silver/user/"${USER}"/multiclass_demo_pipeline"
 training_data : ${prefix}"/training_data"${training_data_version}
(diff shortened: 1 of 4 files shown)


deerzq:
In this case, if people just want to run makeEvaluation not makeTraining, it will not work. I feel perhaps you can have 
`makeTraining(sc, query, isMulticlass).coalesce(numShards, shuffle).map(Util.encode).saveAsTextFile(evalOutput, classOf[GzipCodec])`

in one function, which takes query and output in as input. Then in `makeExampleRun`, you can check if the config has path `training_hive_query`, if so you run the above function for training; then you check if the config has path `eval_hive_query`, if so you run the above function for eval. Then people can run either training or eval or both.

deerzq:
oh, you may want to update https://github.com/airbnb/aerosolve/blob/master/training/src/main/resources/demo_train.conf#L90

and https://github.com/airbnb/aerosolve/blob/master/training/src/main/resources/demo_multiclass_train.conf#L51

deerzq:
And also update the aerosolve hackpad

Patrick1860:
@deerzq good suggestions. I updated it to make it backward compatible as well as updating the demo conf. Will update the hackpad once the new jar is published

deerzq:
LGTM

Patrick1860:
thanks Peng

PR #234, opened 2016-08-03
[demo] fix bug in incomming predict

luanjunyi:
The config for this demo is inconsistent. The used model is `additive_model_with_dynamic_buckets` which specify the output to `${additive_model_name}`. However, eval_testing and eval_training read model from `${spline_model_name}`. This PR fix this bug so that they are consistent.

The `model_output` entry in train_model is not used at all, so it's removed.


deerzq:
LGTM

jq:
👍

PR #235, opened 2016-08-23
Add optional random generator argument to DecisionTreeTrainer methods

yolken:
to: @deerzq @jq 

Content:
- Add optional random generator argument to DecisionTreeTrainer methods; this allows one to make the training deterministic and thus compare the outputs of different pipelines using the same input data


jq:
👍  this is going to fix the compression price curve randomize issue right?

yolken:
@jq Yes, it should.

PR #236, opened 2016-08-23
Add a data conversion job to cast Example to PhotonML Avro format

timyitong:
Given an input of hive table, we would cast it to Example accordingly, applying all transformers possible; and then cast each featureFamily into a separate feature avro array named as:
```
[familyName]Features
```

A special `meta` family is reserved for creating corresponding group ids needed by PhotonML. They will be stored in an Avro ```string -> long``` map. (It's arguable why using longs, but right now seems like stringFeatures casting in Example would make the situation a bit tricky as we also have some implicit delimiter happening). I introduced a flexible trait to allow more sophisticated grouping id extraction (so if we later do want to use string ids instead, we can do that by providing a new `ExampleMetaDataExtractor` implementation), but for now, we are assuming we are extracting all group ids from float features `meta` family.

I added tests for all major components. Also tested on the cluster verified that the job works. 

to: @jq

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -60,6 +60,15 @@ allprojects {
             options.addStringOption('Xdoclint:none', '-quiet')
         }
     }
+
+    // Configure Idea plugin
+    idea {
+        module {
+            scopes.PROVIDED.plus += [configurations.provided]
+
+            downloadJavadoc = true
+        }
+    }
 }
 
 subprojects {
--- a/demo/image_impressionism/build.gradle
+++ b/demo/image_impressionism/build.gradle
@@ -9,6 +9,8 @@ repositories {
 }
 
 dependencies {
+    compile 'org.scala-lang:scala-library:2.10.4'
+
     compile project(':training')
     compile 'org.slf4j:slf4j-simple:1.7.7'
     provided 'org.apache.spark:spark-core_2.10:1.2.1'
--- a/training/build.gradle
+++ b/training/build.gradle
@@ -31,7 +31,16 @@ bintray {
 }
 
 dependencies {
+  compile 'org.scala-lang:scala-library:2.10.4'
   compile project(':core')
+
+  compile ('org.apache.avro:avro:1.7.7') {
+    transitive = false
+  }
+  compile ('org.apache.avro:avro-mapred:1.7.7') {
+    transitive = false
+  }
+
   compile 'com.fasterxml.jackson.module:jackson-module-scala_2.10:2.4.2'
   compile 'joda-time:joda-time:2.5'
   compile 'org.apache.hadoop:hadoop-client:2.2.0'
--- a/training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
@@ -1,0 +1,178 @@
+package com.airbnb.aerosolve.training.photon.ml.data
+
+import java.io.File
+
+import com.airbnb.aerosolve.core.{Example, FeatureVector}
+import com.airbnb.aerosolve.core.transforms.Transformer
+import com.airbnb.aerosolve.training.pipeline.GenericPipeline
+import com.airbnb.aerosolve.training.photon.ml.data.PhotonMLUtils._
+import com.typesafe.config.{Config, ConfigFactory, ConfigParseOptions, ConfigResolveOptions}
+import org.apache.avro.generic.{GenericData, GenericRecord}
+import org.apache.avro.mapred.AvroKey
+import org.apache.avro.mapreduce.{AvroJob, AvroKeyOutputFormat}
+import org.apache.avro.Schema
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.Path
+import org.apache.hadoop.io.NullWritable
+import org.apache.hadoop.mapreduce.Job
+import org.apache.spark.{SparkConf, SparkContext}
+import org.apache.spark.rdd.RDD
+import org.slf4j.LoggerFactory
+
+import scala.collection.JavaConverters._
+import scala.collection.immutable
+import scala.util.Try
+
+
+/**
+  * Providing a data conversion job casting Thrift [[com.airbnb.aerosolve.core.Example]] to PhotonML
+  * Avro format: https://github.com/linkedin/photon-ml#input-data-format
+  *
+  * @see resources/demo_convert_example_to_photonml_avro.conf about how to set up job configuration.
+  *
+  */
+object ConvertExamplesToPhotonMLAvroJob {
+  private val logger = LoggerFactory.getLogger(this.getClass())
+  private val LABEL = "label"
+
+  /**
+    * A single entry-point for launching this job from command line
+    *
+    * @param args Commandline arguments
+    */
+  def main(args: Array[String]): Unit = {
+    val configPath = new File(args(0))
+
+    val config = if (configPath.exists()) {
+      ConfigFactory.load(ConfigFactory.parseFile(configPath))
+    } else {
+      ConfigFactory.load(configPath.getPath(),
+        ConfigParseOptions.defaults.setAllowMissing(false),
+        ConfigResolveOptions.defaults)
+    }
+
+    val sparkConf = new SparkConf().setAppName("Convert Thrift Example to PhotonML Avro")
+    val sc = new SparkContext(sparkConf)
+
+    convertExamplesToPhotonGAMEAvro(sc, config)
+  }
+
+  private def getLabelFromExample(example: Example): Double = {
+    if (example.getExample().get(0).floatFeatures.get(GenericPipeline.LABEL).get("") > 0) {
+      1d
+    } else {
+      0d
+    }
+  }
+
+  /**
+    * This method converts an RDD of examples to PhotonML GAME Avro formats and save to HDFS.
+    * Could be used as an entry-point in programming.
+    *
+    * @param sc         The spark context
+    * @param taskConfig config of the task
+    */
+  def convertExamplesToPhotonGAMEAvro(sc: SparkContext,
+                                      taskConfig: Config): Unit = {
+    val (transformedData, avroSchema) = loadExamples(sc, taskConfig)
+
+    val outputPath = new Path(taskConfig.getString("output_path"))
+
+    val fs = outputPath.getFileSystem(new Configuration())
+    fs.delete(outputPath, true)
+
+    val avroSchemaStr = avroSchema.toString()
+    logger.info(s" Configured output Avro schema: ${avroSchema.toString(true)}")
+
+    val metaDataExtractorClass = Try(taskConfig.getString("meta_data_extractor_class"))
+      .getOrElse(classOf[SingleFloatFamilyMetaExtractor].getName())
+
+    val metaDataExtractorContext = Try(taskConfig
+      .getConfig("meta_data_extractor_context").entrySet().asScala
+      .map { case entry => (entry.getKey(), entry.getValue().toString()) }.toMap
+    ).getOrElse(immutable.Map[String, String]())
+
+    val trainingExampleAvros = transformedData.mapPartitions { case iter =>
+      val schema = new Schema.Parser().parse(avroSchemaStr)
+
+      val metaDataExtractor = Class.forName(metaDataExtractorClass)
+        .newInstance().asInstanceOf[ExampleMetaDataExtractor]
+
+      iter.map { case example =>
+        val record = new GenericData.Record(schema)
+
+        val label = getLabelFromExample(example)
+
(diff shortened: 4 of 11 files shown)


demo/image_impressionism/build.gradle
```diff
@@ -9,6 +9,8 @@ repositories {
 }
 
 dependencies {
+    compile 'org.scala-lang:scala-library:2.10.4'
```
timyitong: These are to make `gradle idea` work more properly.

build.gradle
```diff
@@ -60,6 +60,15 @@ allprojects {
             options.addStringOption('Xdoclint:none', '-quiet')
         }
     }
+
+    // Configure Idea plugin
+    idea {
+        module {
+            scopes.PROVIDED.plus += [configurations.provided]
```
timyitong: Without such config, thrift gen codes won't be viewable in Idea IDE.

demo/image_impressionism/build.gradle
jq: use 2.11？

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
```diff
@@ -0,0 +1,170 @@
...
+
+
+/**
+  * Providing a data conversion job casting Thrift [[com.airbnb.aerosolve.core.Example]] to PhotonML
+  * Avro format: https://github.com/linkedin/photon-ml#input-data-format
+  *
+  */
+object ConvertExamplesToPhotonMLAvroJob {
+
+  import PhotonMLUtils._
```
jq: any reason to put it inside?

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
```diff
@@ -0,0 +1,170 @@
...
+import scala.collection.immutable
+import scala.util.Try
+
+
+/**
+  * Providing a data conversion job casting Thrift [[com.airbnb.aerosolve.core.Example]] to PhotonML
+  * Avro format: https://github.com/linkedin/photon-ml#input-data-format
+  *
+  */
+object ConvertExamplesToPhotonMLAvroJob {
```
jq: add a template conf file and use the generic pipeline to load it and update the generic pipeline doc, it is easier for others to follow.

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
```diff
@@ -0,0 +1,170 @@
...
+
+    // Read and cast to Examples
+    val examples = hiveTraining
+      .map(x => GenericPipeline.hiveTrainingToExample(x, dataFrameSchema, false))
+
+    // Pass through transformers
+    val transformer = new Transformer(taskConfig, "model_config")
+    val transformerBC = sc.broadcast(transformer)
+
+    val transformedExamples = examples.flatMap { case example =>
```
jq: use LinearRankerUtils.makePointwiseFloat?

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  // Load RDD[Example] according to context information
+  private def loadExamples(sc: SparkContext,
+                           taskConfig: Config): (RDD[Example], Schema) = {
+
+    val hiveTraining = GenericPipeline.runQuery(sc, taskConfig.getString("input_hive_query"))
+    val dataFrameSchema = hiveTraining.schema.fields
+
+    val namespaces = (dataFrameSchema.map { case f =>
```
jq: here you get family name as the featureNamespaces for photon?

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
```diff
@@ -0,0 +1,170 @@
...
+      val schema = new Schema.Parser().parse(avroSchemaStr)
+
+      val metaDataExtractor = Class.forName(metaDataExtractorClass)
+        .newInstance().asInstanceOf[ExampleMetaDataExtractor]
+
+      iter.map { case example =>
+        val record = new GenericData.Record(schema)
+
+        // TODO label format should be more robust
+        val label = if (example.getExample().get(0)
```
jq: make it a method  getLabelFromExample

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
timyitong: Yes, Avro schema is very restrictive that we need to declare ahead what are the fields. So this logic is reflecting what are all the feature family names to use.

timyitong:
The CI timed out on trainer tests. I don't think they are related. Any idea what might be the issue?

demo/image_impressionism/build.gradle
timyitong: We are compiling against Spark 2.10, so we shouldn't use 2.11, it will be inconsistent.

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
timyitong: Moving it out

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
timyitong: I think that method is assuming an input of AdditiveModel instance, not sure whether that's general enough. It complicates things that then we need to mock a proper model instance.

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/SingleFloatFamilyMetaExtractor.scala
```diff
@@ -0,0 +1,36 @@
...
+
+    val f = example.getExample().get(0).floatFeatures
+
+    if (!f.containsKey(metaFamily)) {
+      throw new IllegalArgumentException(
+        s"meta data feature field: [${metaFamily}] not presented in " +
+          s"the example's float features: ${example.toString()}")
+    }
+
+    f.get(metaFamily).asScala.foreach { case (name, value) =>
```
jq: so we only allow one meta family, all meta family fields should start with same prefix_?

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
```diff
@@ -0,0 +1,170 @@
...
+    val metaDataExtractorClass = Try(taskConfig.getString("meta_data_extractor_class"))
+      .getOrElse(classOf[SingleFloatFamilyMetaExtractor].getName())
+
+    val metaDataExtractorContext = Try(taskConfig
+      .getConfig("meta_data_extractor_context").entrySet().asScala
+      .map { case entry => (entry.getKey(), entry.getValue().toString()) }.toMap
+    ).getOrElse(immutable.Map[String, String]())
+
+    val trainingExampleAvros = transformedData.mapPartitions { case iter =>
+      val schema = new Schema.Parser().parse(avroSchemaStr)
```
jq: should loadExamples return schema?

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/PhotonMLUtils.scala
```diff
@@ -0,0 +1,198 @@
...
+    r.put(AVRO_VALUE, value)
+    r
+  }
+
+  /**
+    * Returns the avro feature array field name for a certain feature family
+    *
+    * @param familyName
+    */
+  def makeAvroFeaturesFieldName(familyName: String): String = familyName.toLowerCase + AVRO_FEATURES.capitalize
```
jq: so aerosolve family name transfer to familynameFEATURE as the name in avro?

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
```diff
@@ -0,0 +1,170 @@
...
+
+    val namespaces = (dataFrameSchema.map { case f =>
+      val idx = f.name.indexOf("_")
+
+      if (idx == -1) {
+        f.name
+      } else {
+        f.name.substring(0, idx)
+      }
+    }.toSet[String] ++ taskConfig.getStringList("avro_extra_namespaces").asScala).toArray
```
jq: what should be in avro_extra_namespaces?

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
timyitong: This is to prevent Avro schema object serialization issue. As far as I know, Avro internally are keeping some static state information, directly serializing the object is dangerous. So that's why I cast to string, and then deserialize it in remote executors.

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/SingleFloatFamilyMetaExtractor.scala
timyitong: Yes, by default, grouping ids are all stored in fields prefixed with "meta_", but we can also provide a different `context` map to override the settings if other namespace is desired.

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
timyitong: done

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/PhotonMLUtils.scala
timyitong: Yes, but it's actually `familynameFeatures`, `capitalize` will move the 1st char to uppercase.

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
timyitong: Any extra family names generated by transformers but not presented in raw Hive inputs should be added in `avro_extra_namespaces`. We could also call the transformer to reflect such names but I figured this way is probably easier since we are gonna declare those transformers needed in the end anyway.

training/build.gradle
```diff
@@ -31,7 +31,16 @@ bintray {
 }
 
 dependencies {
+  compile 'org.scala-lang:scala-library:2.10.4'
   compile project(':core')
+
+  compile ('org.apache.avro:avro:1.7.7') {
+    transitive = false
```
timyitong: Turns out we have some transitive dependencies conflicts previously (jackson lib probably). It's fine to just exclude transitive here as Avro's dependencies are minimal and they should also be provided by either spark or hadoop.

training/src/main/scala/com/airbnb/aerosolve/training/photon/ml/data/ConvertExamplesToPhotonMLAvroJob.scala
```diff
@@ -0,0 +1,176 @@
...
+    AvroJob.setOutputKeySchema(job, avroSchema)
+    val hadoopConf = job.getConfiguration()
+
+    // Compress Avro binary files
+    hadoopConf.set("mapreduce.output.fileoutputformat.compress", "true")
+    hadoopConf.set(AvroJob.CONF_OUTPUT_CODEC, "deflate")
+
+    // Save as Avro records
+    trainingExampleAvros
+      .map { r => (new AvroKey(r), NullWritable.get()) }
```
jq: so GenericData.Record has to be wrap by AvroKey?  what's the use for NullWritable.get() here?

PR #237, opened 2016-08-26
temp fix of OOM from large group by

luanjunyi:
This PR is temporary fix to an issue where an Spark job that depends on `Evaluation` fail due to out of memory exception.

The root cause was introduced in https://github.com/airbnb/aerosolve/pull/231.  We ran `.collect` on RDD too early, resulting huge list when data is big. This PR won't fix the root cause but will get let the job pass by removing the most expensive operation on the big list. I've verified that the job will successfully finish with this change.

Because the job has been failing two weeks ago, we need some time to backfill. So I propose to merge this PR ASAP so that the failing job will pick up. After that, we can work on the permanent fix .

@jq @deerzq


saurfang:
Oh I had talked with Andy about this before. We should just move all the code to RDD instead of List. I have a fix and will send a PR shortly.

PR #238, opened 2016-08-26
Perform Evaluation on RDD

saurfang:
Revert code clean up that merges the evaluation of records from RDD and List. Instead of merging the code path using List, perform evaluation on RDD instead to handle large dataset.

There is no reliable way to get a SparkContext especially with the way we set up our tests. Since there is really no need to run evaluation on List of records and lists can always be parallelize with single function call, we are removing evaluation on list altogether.

Related to #237 #231 

@jq @deerzq @luanjunyi

### DIFF
--- a/demo/twenty_news/build.gradle
+++ b/demo/twenty_news/build.gradle
@@ -17,11 +17,6 @@ dependencies {
   compile 'org.slf4j:slf4j-simple:1.7.7'
 }
 
-compileScala {
-    scalaCompileOptions.fork = true
-    scalaCompileOptions.forkOptions.jvmArgs = ['-XX:MaxPermSize=512m']
-}
-
 shadowJar {
   zip64 = true
 }
--- a/training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
@@ -1,14 +1,14 @@
 package com.airbnb.aerosolve.training
 
-import org.slf4j.{Logger, LoggerFactory}
-import org.apache.spark.rdd.RDD
 import com.airbnb.aerosolve.core.EvaluationRecord
 import com.airbnb.aerosolve.training.pipeline.ResultUtil
-import org.apache.spark.SparkContext._
-import scala.collection.{mutable, Map}
-import scala.collection.mutable.{ArrayBuffer, Buffer}
-import scala.collection.JavaConversions._
+import org.apache.spark.{SparkConf, SparkContext}
+import org.apache.spark.rdd.RDD
+import org.slf4j.{Logger, LoggerFactory}
+
 import scala.collection.JavaConverters._
+import scala.collection.mutable.ArrayBuffer
+import scala.collection.{Map, mutable}
 
 /*
 * Given an RDD of EvaluationRecord return standard evaluation metrics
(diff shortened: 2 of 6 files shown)


demo/twenty_news/build.gradle
```diff
@@ -17,15 +17,10 @@ dependencies {
   compile 'org.slf4j:slf4j-simple:1.7.7'
 }
 
-compileScala {
```
luanjunyi: Why remove these `compileScala`?

training/src/main/scala/com/airbnb/aerosolve/training/Evaluation.scala
```diff
@@ -92,7 +83,16 @@ object Evaluation {
       .toArray
   }
 
-  def evaluateMulticlassClassification(records : RDD[EvaluationRecord]) : Array[(String, Double)] = {
+  def evaluateBinaryClassificationWithResults(records: List[EvaluationRecord],
+                                              buckets: Int,
+                                              evalMetric: String,
+                                              resultsOutputPath: String): Array[(String, Double)] = {
+    val rdd = SparkContext.getOrCreate().parallelize(records)
+    val result = evaluateBinaryClassificationWithResults(rdd, buckets, evalMetric, resultsOutputPath)
```
luanjunyi: Let's simplify, just one line:

`evaluateBinaryClassificationWithResults(rdd, buckets, evalMetric, resultsOutputPath)`

demo/twenty_news/build.gradle
saurfang: The `fork` option has been removed in Gradle 3+ and is causing error in my local environment. https://docs.gradle.org/current/release-notes#ant-based-scala-compiler-has-been-removed

timyitong: Side discussion, can we switch to gradle wrapper? This could ensure that everyone's using exactly same gradle to compile.

saurfang: @timyitong  +1 if you send a PR ;)

timyitong:
Why local spark context is "not reliable"? Is it due to thread safety issue? One thing I do know is that in local tests, all spark context related methods should be synchronized, local spark context is not thread safe; they may stomp on each other due to parallel test execution.

saurfang:
@timyitong The issue I faced was `SparkContext.getOrCreate` uses a default Spark conf if it is creating a new one. The conf would initialize with system properties and in this case we have to set master and app name explicitly. 
The particular test case on List evaluation wasn't using any SparkContext at all and I don't feel this function should take a SparkContext, which would defeat the design (sort of). Then there was the good old issue where you can't have multiple spark sessions. 
I suppose the correct way to do it is to initialize a SparkContext explicitly in the test case even though it is evaluating on a `List`, then `SparkContext.getOrCreate` will just pick up that. 

Anyway, long story short, I don't feel there is particular value to have the List version so simplifying is probably good enough.

demo/twenty_news/build.gradle
timyitong: Alright, working on this. I assume we are happy with latest Gradle 2.x? Or wanna test out the fresh gradle 3?

saurfang: Don't think anyone here has more experience with gradle than you do ;) I don't have a strong opinion either.

jq:
👍

PR #239, opened 2016-08-26
Remove more sources of uncontrolled randomness from DecisionTreeTrainer

yolken:
to: @jq 

Content:
- Remove more sources of uncontrolled randomness from DecisionTreeTrainer


jq:
👍

PR #240, opened 2016-08-26
Move Json Parser to utils

jq:
@yolken


training/src/test/scala/com/airbnb/aerosolve/training/JsonParserTest.scala
```diff
@@ -1,5 +1,6 @@
 package com.airbnb.aerosolve.training
 
+import com.airbnb.aerosolve.training.utils.JsonParser
 import org.junit.Test
```
saurfang: move this test to utils instead?

jq:
@saurfang Thanks for review, done

saurfang:
👍

PR #241, opened 2016-08-26
Add option to initialize model using quantiles

saurfang:
Using quantiles (e.g. 5th, 95th) instead of min/max to initialie spline can create better buckets for features that are unimodal but heavily skewed.
Broadcast model during feature flattening to minimize task deserialization time (this is crucial for partition sample to work).
Initialize feature as point when min and max is the same for Spline. This will make BIAS a point feature instead of a Spline feature.
A few other code clean up.

@jq @deerzq @timyitong

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/transforms/MoveFloatToStringTransform.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/transforms/MoveFloatToStringTransform.java
@@ -21,6 +21,7 @@ public class MoveFloatToStringTransform implements Transform {
   private String outputName;
   private List<String> keys;
   private double cap;
+  private boolean keep;
 
   @Override
   public void configure(Config config, String key) {
@@ -34,6 +35,11 @@ public class MoveFloatToStringTransform implements Transform {
       cap = config.getDouble(key + ".cap");
     } else {
       cap = 1e10;
+    }
+    if (config.hasPath(key + ".keep")) {
+      keep = config.getBoolean(key + ".keep");
+    } else {
+      keep = false;
     }
   }
 
@@ -56,7 +62,9 @@ public class MoveFloatToStringTransform implements Transform {
     if (keys != null) {
       for (String key : keys) {
         moveFloat(feature1, output, key, cap, bucket);
-        feature1.remove(key);
+        if(!keep) {
+          feature1.remove(key);
+        }
       }
     } else {
       for (Iterator<Entry<String, Double>> iterator = feature1.entrySet().iterator();
@@ -65,7 +73,9 @@ public class MoveFloatToStringTransform implements Transform {
         String key = entry.getKey();
 
         moveFloat(feature1, output, key, cap, bucket);
-        iterator.remove();
+        if(!keep) {
+          iterator.remove();
+        }
       }
     }
   }
--- a/core/src/main/java/com/airbnb/aerosolve/core/transforms/MoveFloatToStringTransform.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/transforms/MoveFloatToStringTransform.java
@@ -13,7 +13,9 @@ import java.util.List;
 
 /**
  * Moves named fields from one family to another. If keys are not specified, all keys are moved
- * from the float family.
+ * from the float family. Features are capped via a `cap` config, which defaults to 1e10, to avoid
+ * exploding string features. The original float feature is removed but can be overridden using
+ * `keep` boolean config.
  */
 public class MoveFloatToStringTransform implements Transform {
   private String fieldName1;
--- a/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
@@ -159,7 +159,8 @@ object AdditiveModelTrainer {
                         linearFeatureFamilies: Set[String],
                         priors: Array[String],
                         minCount: Int,
-                        nDTreePipelineParams: NDTreePipelineParams)
+                        nDTreePipelineParams: NDTreePipelineParams,
+                        initQuantiles: Seq[Double])
 
   case class LearningRateParams(initialLearningRate: Double,
                                 learningRateDecay: Double,
@@ -237,9 +238,11 @@ object AdditiveModelTrainer {
     // transform example as-is to initialize model which we can base our vector compression upon
     val transformed = (frac: Double) => LinearRankerUtils.makePointwiseFloat(input(frac), transformer)
     var model = modelInitialization(sc, transformed, params)
+    var modelBC = sc.broadcast(model.generateFeatureIndexer())
     var bestModel = model
 
-    val Array(trainVectorRDD, validationVectorRDD) = transformFeatureVectorAndSplit(sc, input, params, transformer, model)
+    // broadcast a separate copy of model because the modelBC will be destroyed after each iteration
+    val Array(trainVectorRDD, validationVectorRDD) = transformFeatureVectorAndSplit(sc, input, params, transformer, sc.broadcast(model))
 
     var learningRate = params.learningRate.initialLearningRate
     var trainingLosses = List(Double.MaxValue)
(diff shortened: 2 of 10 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -159,7 +159,8 @@ object AdditiveModelTrainer {
                         linearFeatureFamilies: Set[String],
                         priors: Array[String],
                         minCount: Int,
-                        nDTreePipelineParams: NDTreePipelineParams)
+                        nDTreePipelineParams: NDTreePipelineParams,
+                        initQuantiles: Seq[Double])
```
jq: Nice, can you add the new quantiles to one of the demo conf?

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -237,21 +238,22 @@ object AdditiveModelTrainer {
...
     var validationLosses = List(Double.MaxValue)
     var bestValidationLoss = Double.MaxValue
     var bestIteration = 0
     val sgdParams = SgdParams(params, sc.accumulator(0), sc.accumulator(0))
-    var modelBC = sc.broadcast(model.generateFeatureIndexer())
     var i = 0
     while (i < iterations &&
       (params.loss.earlyStopping == 0 || (i - bestIteration) < params.loss.earlyStopping) &&
-      (trainingLosses.length < 2 || (trainingLosses(1) - trainingLosses(0) > params.loss.convergenceTolerance))
+      (trainingLosses.length < 2 || params.loss.convergenceTolerance <= 0 || (trainingLosses(1) - trainingLosses(0) > params.loss.convergenceTolerance))
```
jq: params.loss.convergenceTolerance <= 0  what case we set it to <= 0?

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
```diff
@@ -237,21 +238,22 @@ object AdditiveModelTrainer {
     // transform example as-is to initialize model which we can base our vector compression upon
     val transformed = (frac: Double) => LinearRankerUtils.makePointwiseFloat(input(frac), transformer)
     var model = modelInitialization(sc, transformed, params)
+    var modelBC = sc.broadcast(model.generateFeatureIndexer())
     var bestModel = model
 
-    val Array(trainVectorRDD, validationVectorRDD) = transformFeatureVectorAndSplit(sc, input, params, transformer, model)
+    // broadcast a separate copy of model because the modelBC will be destroyed after each iteration
+    val Array(trainVectorRDD, validationVectorRDD) = transformFeatureVectorAndSplit(sc, input, params, transformer, sc.broadcast(model))
```
jq: should it change interface to Broadcast[]?

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
```diff
@@ -256,17 +265,32 @@ object TrainingUtils {
...
+          // TODO: use built-in percentileApprox in Spark 2.0
+          quantiles.map(q => callUDF("percentile_approx", $"value", lit(q))): _*
       )
       // ignore features present in less than minCount examples
       .where($"count" >= minCount)
       .map {
-        case Row(family: String, feature: String, count: Long, min: Double, max: Double, sum: Double, sqSum: Double) =>
-          ((family, feature), FeatureStatistics(count, min, max, sum / count, (sqSum - sum * sum / count) / (count - 1)))
+        row =>
+          val Seq(family: String, feature: String, count: Long, min: Double, max: Double, sum: Double, sqSum: Double) = row.toSeq.take(7)
```
jq: use a const for both 7?

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
```diff
@@ -256,17 +265,32 @@ object TrainingUtils {
...
-        sum($"value" * $"value")
+        Seq(
+          min($"value"),
+          max($"value"),
+          sum($"value"),
+          // TODO: use built-in stdev function in Spark 1.6
+          sum($"value" * $"value")
+        ) ++
+          // TODO: use built-in percentileApprox in Spark 2.0
+          quantiles.map(q => callUDF("percentile_approx", $"value", lit(q))): _*
```
jq: what's the magic of : _* here?

jq:
some unit test failed

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
saurfang: Good call. Added.

training/src/main/scala/com/airbnb/aerosolve/training/AdditiveModelTrainer.scala
saurfang: When you don't want to stop iteration even when training loss might be increasing. This can happen if training and evaluation loss is quite correlated. More iterations might still help because learning rate decay will kick in and could possibly further decrease loss down the road. Those cases are much better addressed using early stopping, which stops iterations early when loss has not improved for certain number of iterations. `convergenceTolerance` is useful if you want to experiment things quickly to get a "good enough" model.

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
saurfang: addressed

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
saurfang: It's a type casting. The function takes `columns: Column*` for argument. What we have is `Seq[Column]`. Normally you would do `columns: Column*` but in this case, you can just do `columns: _*` since `Column` type can be easily inferred. http://stackoverflow.com/questions/6051302/what-does-colon-underscore-star-do-in-scala

core/src/main/java/com/airbnb/aerosolve/core/transforms/MoveFloatToStringTransform.java
```diff
@@ -21,6 +21,7 @@
   private String outputName;
   private List<String> keys;
   private double cap;
+  private boolean keep;
```
deerzq: edit the comments in line 15 - 16 to address this change.

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
```diff
@@ -226,16 +227,33 @@ object TrainingUtils {
...
+                              )
 
   /**
     * Compute [[FeatureStatistics]] across examples and ignore those with less than minCount features.
     */
-  def getFeatureStatistics(minCount: Int, input: RDD[Example]): Array[((String, String), FeatureStatistics)] = {
-    val sqlContext = SQLContext.getOrCreate(input.sparkContext)
+  def getFeatureStatistics(minCount: Int,
+                           input: RDD[Example],
+                           quantiles: Seq[Double] = Nil): Array[((String, String), FeatureStatistics)] = {
```
deerzq: it seems the length of  `quantiles` should be 2, so perhaps adding a check for the length?

deerzq:
some nits, LGTM

core/src/main/java/com/airbnb/aerosolve/core/transforms/MoveFloatToStringTransform.java
saurfang: added. thx

training/src/main/scala/com/airbnb/aerosolve/training/TrainingUtils.scala
saurfang: This was checked in additive trainer instead because it makes sense to gather more quantiles out of the context of additive trainer. When we have `approxPercentile` in Spark 2.0, we will probably be able to afford to get a lot more quantiles and do a more efficient dynamic bucketing.

PR #242, opened 2016-08-29
Add gradle wrapper to repo

timyitong:
Gradle wrapper is a more stable and env-independent gradle that allows multiple
gradle versions work with different projects and avoid stomping upon each other. 

According to gradle doc: https://docs.gradle.org/current/userguide/gradle_wrapper.html

Three things needed to be checked in:
```
gradlew
gradlew.bat
gradle/
```

But after we checked in them once, next time we only need to bump up gradle
versions in `gradle/wrapper/gradle-wrapper.properties` to get latest gradle.

For people who are a bit annoyed by `gradlew` because you need to retrieve back to project root to call the script. Here is a cool script redirecting `gradle` to use `gradlew` if one existed in a parent directory: https://github.com/cbeams-archive/shell-scripts ; and fall back to system `gradle` if none exists. 

@jq @saurfang @deerzq


saurfang:
:shipit:

PR #243, opened 2016-08-30
Evaluation on List and RDD

saurfang:
It turns out evaluation on list is indeed useful when we want to perform them on bucketed dataset. Those bucketed results are often small enough that they can run on executor in parallel. Adding List based evaluation back for binary classification with a little bit of suboptimal DRY.

@luanjunyi @deerzq @jq


deerzq:
LGTM

luanjunyi:
@saurfang ++

PR #244, opened 2016-08-30
Use Gradle Wrapper in Travis

saurfang:
Use `gradlew` instead of `gradle` so wrapper is correctly used and tested.

@timyitong


timyitong:
LGTM

PR #245, opened 2016-09-01
Float Label Transform

saurfang:
Add a new transform that creates LABEL feature based on a Float feature with a threshold value. With custom merge strategy, one can combine several float features into a single LABEL.

I also considered adding a LabelStrategy/Comparator/Ordering configuration, which would dictate what constitutes a positive label. However I couldn't find an out of the box enum that does the job (i.e. GTE, GT, EQ, NEQ, LT, LTE). I then feel like it is probably an overkill and we could just have a negate float feature transform to handle a naive cases. Or rephrase the model problem by swapping the definition of positive/negative label.

@yinlou @luanjunyi @timyitong @deerzq @jq

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatLabelTransform.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatLabelTransform.java
@@ -1,0 +1,69 @@
+package com.airbnb.aerosolve.core.transforms;
+
+import com.airbnb.aerosolve.core.FeatureVector;
+import com.airbnb.aerosolve.core.util.Util;
+
+import java.util.Map;
+
+import com.typesafe.config.Config;
+
+/**
+ * Convert a Float feature to binary LABEL based on threshold value. This is intended for
+ * binary classification where LABEL is either -1 or 1. MergeStrategy describe how existing
+ * LABEL is treated.
+ *
+ * The purpose of MergeStrategy is to help combining float features into LABEL. For example,
+ * using OVERRIDE_NEGATIVE is equivalent to LABEL with feature1 >= threshold1 or feature2 >= threshold2,
+ * using OVERRIDE_POSITIVE is equivalent to LABEL with feature1 >= threshold1 and feature2 >= threshold2.
+ */
+public class FloatLabelTransform implements Transform {
+  enum MergeStrategy {
+    OVERRIDE, // override existing label
+    OVERRIDE_NEGATIVE, // override negative label but keep positive label
+    OVERRIDE_POSITIVE, // override positive label but keep negative label
+    SKIP // preserve existing label (only replace if observation is un-labeled)
+  }
+
+  private String fieldName1;
+  private String key1;
+  private double threshold;
+  private MergeStrategy mergeStrategy;
+
+  @Override
+  public void configure(Config config, String key) {
+    fieldName1 = config.getString(key + ".field1");
+    key1 = config.getString(key + ".key1");
+    threshold = config.getDouble(key + ".threshold");
+    mergeStrategy = MergeStrategy.valueOf(config.getString(key + ".merge").toUpperCase());
+  }
+
+  @Override
+  public void doTransform(FeatureVector featureVector) {
+    Map<String, Double> labelFeature = Util.getOrCreateFloatFeature("LABEL", featureVector.floatFeatures);
+
+    Double label = labelFeature.get("");
+
+    if (mergeStrategy == MergeStrategy.SKIP && label != null) return;
+
+    Map<String, Double> floatFeature = featureVector.floatFeatures.get(fieldName1);
+    if (floatFeature != null) {
+      Double featureValue = floatFeature.get(key1);
+      if (featureValue != null) {
+        double newLabel = featureValue >= threshold ? 1 : -1;
+
+        if (
+          // fill in missing label
+            label == null ||
+                // ignore existing label
+                mergeStrategy == MergeStrategy.OVERRIDE ||
+                // override negative label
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatLabelTransform.java
```diff
@@ -0,0 +1,65 @@
...
+
+    if (mergeStrategy == MergeStrategy.KEEP_ALL && label != null) return;
+
+    Map<String, Double> floatFeature = featureVector.floatFeatures.get(fieldName1);
+    if (floatFeature != null) {
+      Double featureValue = floatFeature.get(key1);
+      if (featureValue != null) {
+        double newLabel = featureValue >= threshold ? 1 : -1;
+
+        if (
```
luanjunyi: NIT: for better readability, how about refact as

```scala
val override_missing = label == null
val override_existing = mergeStrategy == MergeStrategy.IGNORE 
val override_negative = (mergeStrategy == MergeStrategy.KEEP_POSITIVE && label < 0 && newLabel > 0) 
val override_positive = (mergeStrategy == MergeStrategy.KEEP_NEGATIVE && label > 0 && newLabel < 0)

if (List(override_missing, override_existing, override_negative, override_positive).exists(_ == true)) {
  labelFeature.put("", newLabel);
}

```

luanjunyi:
It will be great if you can add in the comment to demo how to `With custom merge strategy, one can combine several float features into a single LABEL`.

LGTM except for a nitpicking

core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatLabelTransform.java
```diff
@@ -0,0 +1,65 @@
...
+import com.typesafe.config.Config;
+
+/**
+ * Convert a Float feature to binary LABEL based on threshold value. This is intended for
+ * binary classification where LABEL is either -1 or 1. MergeStrategy describe how existing
+ * LABEL is treated.
+ */
+public class FloatLabelTransform implements Transform {
+  enum MergeStrategy {
+    IGNORE, // ignore existing label
```
luanjunyi: How about 'OVERRIDE' instead of 'IGNORE'?

saurfang: Good idea

core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatLabelTransform.java
saurfang: Transform is performance critical so taking advantage of short circuiting is very important here. Same goes for using ENUM over String, which I believe we have a few places in Aerosolve that can use such improvements.

I could break four conditions into local variables but it is unclear whether compiler will inline those, JIT will optimize during runtime, or short circuiting will not apply.

saurfang:
Added more comments to give some intuition regarding combing features. I am not a big fan of the transform name itself "FloatLabelTransform". Looking for ideas otherwise I'll merge it.

core/src/main/java/com/airbnb/aerosolve/core/transforms/FloatLabelTransform.java
luanjunyi: I don't insist on changing it to separate variables. But I would like to share my experience on this kind of 'optimization'. Many people are taught about this kind of short-circuit could produce big performance gain. But in practice, it never worked.

There are many reasons behind. But in general, it's because any program has much larger overhead to run which usually involve many I/O. That overhead will marginalize the performance gain from small memory operation.

For this particular case, I actually ran an experiment to demonstrate this effect. I think the two programs should have more pronounced performance difference from the short-circuiting than our real spark jobs. 

The short-circuiting version:

```scala
object ShortCut {
  def main(args: Array[String]) {
    var count = 0
    val r = scala.util.Random
    for (i <- 1 to 1000) {
      if (r.nextDouble >= 0.5 || r.nextDouble >= 0.5 || r.nextDouble >= 0.5 || r.nextDouble >= 0.5) {
        count = count + 1
      }
    }
  }
}
```

The non-short-circuiting version:

```scala
object NonShortCut {
  def main(args: Array[String]) {
    var count = 0
    val r = scala.util.Random
    for (i <- 1 to 1000) {
      val a = r.nextDouble
      val b = r.nextDouble
      val c = r.nextDouble
      val d = r.nextDouble
      //println("%.3f, %.3f, %.3f, %.3f".format(a, b, c, d))
      if (a >= 0.5 || b >= 0.5 || c >= 0.5 || d >= 0.5) {
        count = count + 1
      }
    }
  }
}
```

Compiling and run each ten times, the elapsed time are almost always the same. Below is one instance of the experiment.

```
 $ time for i in `seq 1 10`; do scala NonShortCut; done
941
951
932
943
938
928
943
928
928
930

real	0m3.284s
user	0m3.117s
sys	0m0.976s

 $ time for i in `seq 1 10`; do scala ShortCut; done
947
945
928
939
938
940
950
942
941
942

real	0m3.300s
user	0m3.137s
sys	0m0.973s

```

Similar proof can be found very widely. For example, Python is considered slow(indeed in lots of context), but training with Sklearn is very fast because all key computation like matrix multiply is done in Fortune or C. The performance issue in user program, like the Python for-loop or branching, have no effect in any real program.

My two cents.

luanjunyi: I agree your profiling is more reasonable in that it excluded JVM starting up time. I think it's your judgment call whether to go for the cleaner code or the 20% performance gain in one transformer. It's not clear how much it will eventually contribute in our production jobs.

PR #246, opened 2016-09-01
Custom Serialization of Model

saurfang:
Enable model that is not implemented in aerosolve-core be serialized and deserialized via ModelFactory. This is achieved by storing Model class name as the model name in model header. This makes it possible to implemented AbstractModel interface in non-aerosolve code. For example, this can support wrapping spark.ml, PhotonML and other models from different libraries.

@timyitong @jq

### DIFF
--- a/core/src/main/java/com/airbnb/aerosolve/core/models/ModelFactory.java
+++ b/core/src/main/java/com/airbnb/aerosolve/core/models/ModelFactory.java
@@ -31,9 +31,15 @@ public final class ModelFactory {
       case "full_rank_linear" : return new FullRankLinearModel();
       case "low_rank_linear" : return new LowRankLinearModel();
       case "multilayer_perceptron" : return new MlpModel();
+      default:
+        log.info("Attempting to initialize " + name);
+        try {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/airbnb/aerosolve/core/models/ModelFactory.java
```diff
@@ -31,9 +31,14 @@ public static AbstractModel createByName(String name) {
       case "full_rank_linear" : return new FullRankLinearModel();
       case "low_rank_linear" : return new LowRankLinearModel();
       case "multilayer_perceptron" : return new MlpModel();
+      default:
+        log.info("Attempting to initialize " + name);
+        try {
+          return (AbstractModel) Class.forName(name).newInstance();
+        } catch (Exception e) {
+          throw new RuntimeException(e);
```
timyitong: Also a detailed error log? `Cannot find customized model class ...bla`?

timyitong:
LGTM other than the logging nitpick.

saurfang:
Addressed. Thanks!

PR #247, opened 2016-10-24
fix closure bug

luanjunyi:
The closure counting minScore and maxScore won't work. This is not very serious but will make the AUC calculation inaccurate. This PR fixes this bug and will force the job to fail if minScore and maxScore equals since something must be deeply wrong and we should fail as loudly as possible.

@deerzq @saurfang


saurfang:
Huh...good catch. LGTM

deerzq:
LGTM

PR #254, opened 2016-11-18
remove used code

luanjunyi:
I came across this code while debugging another issue. Did a code search and found out no one is using it. Besides, it's quite confusing since it assumes the label is stored in feature "$rank". So I propose to remove this function in this PR.

@deerzq


saurfang:
👍

deerzq:
LGTM

PR #255, opened 2016-11-21
shuffle before each SGD epoch

luanjunyi:
The calibration SGD was wrong because it's not shuffling between each epoch. SGD still works but this could cause problems for more unforgiving dataset, let the highly screwed data I used for short-term labeling.

This PR adds the missing random shuffling step to the SGD of Platts Scaling. Let me know if you guys knows better way of shuffling.

@saurfang @deerzq

### DIFF
--- a/training/src/main/scala/com/airbnb/aerosolve/training/ScoreCalibrator.scala
+++ b/training/src/main/scala/com/airbnb/aerosolve/training/ScoreCalibrator.scala
@@ -41,19 +41,18 @@ object ScoreCalibrator {
     var iter = 0
     var old_offset = 0.0
     var old_slope = 1.0
-    val partitionedInput = input.repartition(numBags).cache()
+
     do {
       old_offset = params(0)
       old_slope = params(1)
-      params = trainMiniSGD(partitionedInput, numBags, learningRate, params(0), params(1))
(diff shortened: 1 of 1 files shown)


training/src/main/scala/com/airbnb/aerosolve/training/ScoreCalibrator.scala
```diff
@@ -92,8 +90,12 @@ object ScoreCalibrator {
                    learningRate : Double,
                    offset : Double,
                    slope : Double) : Array[Double] = {
-
-    val result = input
+    // First shuffle the data otherwise SGD won't work.
+    val data = input.map(x => (scala.util.Random.nextInt, x))
+      .sortByKey()
+      .map({case (r, x) => x})
+      .repartition(numBags)
```
saurfang: I don't believe this would perform the shuffling correctly because `repartition` uses hash partitioning by default IIRC. Hence it will give you the same deterministic repartition based on the has key of `(Double, Double)`. I think you can do `input.map(x => (math.rand, x)).sortBy(_._1, numPartitions  = numBags).values` instead.

Also is input cached by default? You may want to add that to avoid repeated I/O.

luanjunyi: Made the change according to our chat. PTAL

PR #260, opened 2016-12-13
Make Point Feature Fixed

saurfang:
Instead of assuming value always is 1, make Point feature always output weight instead. Otherwise if we encounter features that we have not seen, we would make this an effective f(x)=w*x linear feature instead.

This is consistent with the way we use Point feature in AdditiveTrainer where Point is only provisioned when `stats.min == stats.max`.

@jq @deerzq


deerzq:
LGTM @jq please double-check if this will result in any inconsistency for the edgar

jq:
LGTM

PR #263, opened 2017-03-15
Open src strategy model - first step

jq:
first step to open src strategy model.

@saurfang @timyitong @yinlou


mLewisLogic:
At a high level I actually worry about exposing this code beyond our team. The `conf` param hierarchy is pretty messy (common for code that's been iteratively added to). This PR essentially exposes our current config API as "the way" to do things. If anybody outside of Pricing adopted this code, it would limit our ability to make improvements to the way conf files are structured. I feel like we need to get at least the interface boundaries polished before we start promoting other teams onto it or it will create a lot of future work for us, or just prevent us from changing things.

PR #264, opened 2017-05-15
Open src airlearner including strategy model and xgboost pipeline

jq:
@saurfang @mLewisLogic @jieyingchen


airlearner/REVIEWERS
```diff
@@ -0,0 +1 @@
+jq
```
saurfang: Drop this file?

PR #276, opened 2017-07-23
Pick correct FS from path

jq:
Pick correct FS from the path. 
IF default FS is different than the FS in the path, pick the FS in the path


deerzq:
LGTM, thanks for the quick fix

PR #278, opened 2017-08-03
Move predictionLower and predictionIncrease to TrainingSample so that…

jq:
Move predictionLower and predictionIncrease to TrainingSample so that label == true and label == false can be computed differently

@mLewisLogic @deerzq

### DIFF
--- a/airlearner/airlearner-strategy/src/main/scala/com/airbnb/common/ml/strategy/data/BinaryTrainingSample.scala
+++ b/airlearner/airlearner-strategy/src/main/scala/com/airbnb/common/ml/strategy/data/BinaryTrainingSample.scala
@@ -1,6 +1,7 @@
 package com.airbnb.common.ml.strategy.data
 
 import com.airbnb.common.ml.strategy.config.TrainingOptions
+import com.airbnb.common.ml.strategy.eval.BinaryMetrics
 
 
 /*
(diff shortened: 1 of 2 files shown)


airlearner/airlearner-strategy/src/main/scala/com/airbnb/common/ml/strategy/data/BinaryTrainingSample.scala
```diff
@@ -116,4 +117,22 @@ trait BinaryTrainingSample extends BinaryScoringSample {
     // todo test with trueValue
     bound * basePivot
   }
+
+  // TODO maybe need predictionHigher?
```
deerzq: no need to have this?

airlearner/airlearner-strategy/src/main/scala/com/airbnb/common/ml/strategy/data/BinaryTrainingSample.scala
```diff
@@ -116,4 +117,22 @@ trait BinaryTrainingSample extends BinaryScoringSample {
...
+  def predictionLower(prediction: Double): Boolean = {
+    if (label && trueValue.isDefined) {
```
deerzq: when will we have label = True but trueValue undefined?

deerzq:
LGTM

mLewisLogic:
👍

PR #279, opened 2017-08-15
Make xgboost stable param configurable

jq:
Make xgboost stable param configurable, so that we extend it to allow different obj function

@mLewisLogic

### DIFF
--- a/airlearner/airlearner-strategy/src/main/scala/com/airbnb/common/ml/strategy/config/TrainingOptions.scala
+++ b/airlearner/airlearner-strategy/src/main/scala/com/airbnb/common/ml/strategy/config/TrainingOptions.scala
@@ -1,10 +1,10 @@
 package com.airbnb.common.ml.strategy.config
 
+import com.airbnb.common.ml.util.ConfigUtils
 import com.typesafe.config.Config
+
 import scala.collection.JavaConverters._
 import scala.util.Try
-
-import com.airbnb.common.ml.util.ConfigUtils
 
 
 /**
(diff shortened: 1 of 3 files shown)


airlearner/airlearner-strategy/src/main/scala/com/airbnb/common/ml/strategy/config/TrainingOptions.scala
```diff
@@ -1,11 +1,11 @@
 package com.airbnb.common.ml.strategy.config
 
+import com.airbnb.common.ml.util.ConfigUtils
```
mLewisLogic: Why move this import?

airlearner/airlearner-utils/src/main/scala/com/airbnb/common/ml/util/ConfigUtils.scala
```diff
@@ -2,9 +2,10 @@ package com.airbnb.common.ml.util
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable.Map
-
 import com.typesafe.config.Config
 
+import scala.util.Try
```
mLewisLogic: Probably want this with the other `scala.` imports.

airlearner/airlearner-strategy/src/main/scala/com/airbnb/common/ml/strategy/config/TrainingOptions.scala
jq: seems done by IntelliJ, maybe sort by order?

PR #281, opened 2017-08-17
Add xgboost scoring option without updating Hive tables

TDDFT:
Add an option to make Hive partition update optional.

@jq @jieyingchen


jq:
LGTM

PR #282, opened 2017-08-17
make loss configurable 

jq:
make loss configurable so that we support many different types of loss in params search

@mLewisLogic

### DIFF
--- a/airlearner/airlearner-xgboost/src/main/scala/com/airbnb/common/ml/xgboost/model/XGBoostModel.scala
+++ b/airlearner/airlearner-xgboost/src/main/scala/com/airbnb/common/ml/xgboost/model/XGBoostModel.scala
@@ -1,31 +1,21 @@
 package com.airbnb.common.ml.xgboost.model
 
-import scala.collection.mutable
-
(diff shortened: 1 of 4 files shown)


airlearner/airlearner-xgboost/src/main/scala/com/airbnb/common/ml/xgboost/model/XGBoostModel.scala
```diff
@@ -36,22 +26,67 @@ class XGBoostModel(training: DMatrix, eval: DMatrix)
...
+      eval: String,
+      loss: String
+  ): XGBoostModel = {
+    getModelByFileWithLoss(training, eval, getLossFunction(loss))
+  }
+
+  private def getLossFunction(loss: String): (Double, Double) => Double = {
+    val lossFun: (Double, Double) => Double = loss match {
+      case "linear" => XGBoostModel.linearLoss
+      case _ => XGBoostModel.logLoss
```
mLewisLogic: Maybe push the default up to where it's pulled out of the config, and force this method's param to be a valid name?

jq:
@mLewisLogic thanks for review, updated with forcing known loss type

PR #283, opened 2017-08-23
[aerosolve] Drop hive partition before writing to location for extern…

lkhamsurenl:
…al table

### What is the proposed change?

If overwrite option is true, we first need to drop the existing partition from hive otherwise ` HiveUtil.updateHivePartition` will drop just committed new text files. 

### How was this tested?

- [x] Was the change tested with treehouse? 

image

### Reviewers

- @jq @saurfang @mLewisLogic

### DIFF
--- a/airlearner/airlearner-utils/src/main/scala/com/airbnb/common/ml/util/HiveUtil.scala
+++ b/airlearner/airlearner-utils/src/main/scala/com/airbnb/common/ml/util/HiveUtil.scala
@@ -4,14 +4,53 @@ import scala.reflect.ClassTag
 import scala.util.Try
 
 import com.typesafe.config.Config
-import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.fs.Path
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.hive.HiveContext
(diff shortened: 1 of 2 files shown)


airlearner/airlearner-utils/src/main/scala/com/airbnb/common/ml/util/HiveUtil.scala
```diff
@@ -4,15 +4,54 @@ import scala.reflect.ClassTag
...
+    }
+
+    // Break the table name into namespace.table_name
+    val Array(namespace, table) = hiveTableName.split('.')
+    // Turn the partition spec map into a Hive-format String
+    val partitionSpec = hivePartitionSpecsMapToString(hivePartitionSpecs)
+
+    // Drop the partition if it exists
+    hc.sql(s"USE $namespace")
+    hc.sql(s"ALTER TABLE $table DROP IF EXISTS PARTITION ($partitionSpec)")
```
saurfang: This function is just `dropHivePartition(hc, hiveTableName, hivePartitionSpecsMapToString(hivePartitionSpecs))`?

lkhamsurenl:
@saurfang PTAL

